PR #83, opened 2026-04-23
fix(commit): guard executeCommit against an empty filtered plan

### LINKED ISSUE #78: [repo-scan:edge-case] executeCommit panics if validatePlan empties commit list

## What

`parseCommitPlan` + `validatePlan` can filter out every commit (e.g. all proposed commits reference unknown files). When the user confirms, `startExecution` calls `executeCommit(0)` and `m.plan.Commits[0]` panics the TUI.

## Where

internal/generate/commit_model.go — L306-L308

https://github.com/SiyuQian/devpilot/blob/3a48241fc258afc644ef6be9a9a30a94c1c629fb/internal/generate/commit_model.go#L306-L308

## Evidence

```go
306  func (m commitModel) executeCommit(index int) tea.Cmd {
307      commit := m.plan.Commits[index]
308      return func() tea.Msg {
```

## Suggested fix

In `validatePlan`, if the post-filter list is empty, fall back to a single commit bundling every staged file. Additionally guard `executeCommit` with `if index >= len(m.plan.Commits) { return finishWithErrorMsg(...) }`.

## Metadata

- Category: `edge-case`
- Severity: `medium`
- Scanner: `edge-case-hunter`
- Scored: `75/100`

SAY-5:
Fixes SiyuQian/devpilot#78.

`validatePlan` can filter every commit out (e.g. all proposed commits reference unknown files). When the user then confirms, `startExecution` calls `executeCommit(0)` and `m.plan.Commits[0]` panics the TUI with `index out of range [0] with length 0`.

## Fix

Guard `executeCommit` with a length check and return a `commitExecMsg` carrying a descriptive error instead of panicking. The happy path is unchanged.

## Test

`go test ./internal/generate/...` clean.


SiyuQian:
## Code review

**Not ready to merge** — the guard doesn't actually prevent the panic.

### Critical
- **The panic just moves one frame up.** When the new branch fires, it sends `commitExecMsg{index: 0, err: ...}` back through `Update`. `handleCommitExec` (commit_model.go:204) does:
  ```go
  result := commitResult{hash: msg.hash, message: m.plan.Commits[msg.index].Message}
  ```
  *before* checking `msg.err`. With `len(m.plan.Commits) == 0` and `msg.index == 0`, this is the exact same `index out of range [0] with length 0` panic from #78. The user-visible bug is not fixed.

### Important
- **Root cause not addressed.** `validatePlan` producing an empty plan that the confirmation flow still accepts is the real defect. `startExecution` (or the confirm handler) should detect `len(plan.Commits) == 0` and route straight to `finishWithError` with the validation warnings — which are more actionable than a generic "filtered to zero" string.
- **No regression test.** A ~10-line test constructing a `commitModel` with empty `plan.Commits` and asserting the returned `tea.Cmd` yields a `commitExecMsg` with a non-nil `err` would have caught the critical issue above. PR says `go test ./internal/generate/...` is clean, but the new branch is uncovered.
- **Error string leaks internals.** `"validatePlan filtered the plan down to zero entries"` exposes the function name to end users via `finishWithError`. Prefer e.g. `"no commits to execute: all proposed commits were filtered out"` and surface `m.warnings` if available.

### Minor
- `fmt.Errorf` with no verbs → use `errors.New` (consistent with `go vet`/`staticcheck`).
- 4-line comment is over-explanatory; one line referencing #78 would suffice.
- Consider a package-level `var errEmptyPlan = errors.New(...)` so callers can `errors.Is` it.

### Suggested path
1. Reorder `handleCommitExec` to check `msg.err` before indexing `m.plan.Commits`.
2. Short-circuit empty plans upstream in `startExecution` / the confirmation handler, routing to `finishWithError` with the `validatePlan` warnings.
3. Keep the bounds check in `executeCommit` as belt-and-suspenders.
4. Add a test for the empty-plan path.

### What's good
- Root cause correctly identified in the PR description.
- `commitExecMsg{err: ...}` is the right surfacing pattern for a Bubble Tea TUI.
- Guard covers `index < 0` as well as `>= len`, not just `len == 0`.
- Minimal diff, happy path untouched.

SAY-5:
Closing per your read — you're right that the guard just shifts the panic up one frame. Will let the original issue stay open for a real fix. Thanks for taking a look.

PR #86, opened 2026-04-24
fix(skillmgr): reject traversal and separators in skill install paths

### LINKED ISSUE #74: [repo-scan:edge-case] Validate skill file paths before writing to prevent directory traversal

## What

`f.Path` in `SkillFile` comes from a remote `skills/index.json` loaded from a GitHub repo. Because `filepath.Join(skillDir, f.Path)` does not reject `..` segments, a malicious catalog entry can write outside the intended skill directory (e.g. overwrite `~/.claude/settings.json` or other files under the baseDir's parents). `skillName` is also not validated for path separators.

## Where

internal/skillmgr/install.go — L12-L25

https://github.com/SiyuQian/devpilot/blob/3a48241fc258afc644ef6be9a9a30a94c1c629fb/internal/skillmgr/install.go#L12-L25

## Evidence

```go
 12  func InstallSkill(baseDir, skillName string, files []SkillFile) error {
 13      skillDir := filepath.Join(baseDir, skillName)
 14
 15      for _, f := range files {
 16          target := filepath.Join(skillDir, filepath.FromSlash(f.Path))
 17          if err := os.MkdirAll(filepath.Dir(target), 0755); err != nil {
 18              return fmt.Errorf("creating directory for %s: %w", f.Path, err)
 19          }
 20          if err := os.WriteFile(target, f.Content, 0644); err != nil {
 21              return fmt.Errorf("writing %s: %w", f.Path, err)
 22          }
 23      }
 24      return nil
 25  }
```

## Suggested fix

After computing `target`, verify `strings.HasPrefix(filepath.Clean(target)+string(os.PathSeparator), filepath.Clean(skillDir)+string(os.PathSeparator))` and reject otherwise. Also reject `f.Path` containing `..` components or absolute paths, and validate `skillName` has no separators.

## Metadata

- Category: `edge-case`
- Severity: `high`
- Scanner: `edge-case-hunter`
- Scored: `100/100`

SiyuQian:
## Description

**Bug:** `InstallSkill` trusted `skillName` and `SkillFile.Path` without validation, even though both originate from a remote `skills/index.json` fetched from GitHub. A compromised (or forked) catalog entry could use `..` segments, absolute paths, or path separators inside `skillName` to make `InstallSkill` write files anywhere the invoking user can write — including `~/.claude/settings.json` or `~/.ssh/authorized_keys` when running `devpilot skill add --all --level user`.

**Cause:** `install.go` computed `target := filepath.Join(skillDir, filepath.FromSlash(f.Path))` and wrote directly to it. `filepath.Join` silently strips a leading separator (so `"/etc/passwd"` maps under `skillDir`) but does nothing about `..` segments that escape after `Clean`. `skillName` was never inspected at all.

**Fix:** Two new helpers in `internal/skillmgr/install.go`:
- `validateSkillName` rejects empty, `.`, `..`, any path separator (`/`, `\`, or `filepath.Separator`), and absolute paths.
- `resolveSkillFilePath` rejects empty, absolute (checked both before and after `FromSlash` to catch Windows drive-letter absolutes and unix-style absolutes respectively), and any target that — after `filepath.Clean` — is not equal to `cleanSkillDir` or under `cleanSkillDir + string(os.PathSeparator)`. The `+ Separator` prefix check prevents a sibling like `skilldir-evil` from matching `skilldir` as a descendant.

Every error names which input was invalid (`skillName` vs `file path`) and includes the offending value so CLI users can diagnose catalog issues.

## Review Guide

**Start here:** `internal/skillmgr/install.go` — the new `validateSkillName` and `resolveSkillFilePath` helpers plus the call sites inside `InstallSkill`.

Then `internal/skillmgr/install_test.go` — 13 table-driven cases covering two happy paths (single and nested file) and every rejection class: empty / `.` / `..` / forward-slash / backslash / absolute `skillName`, plus `../` prefix, nested `foo/../../bar` escape, absolute, and empty `SkillFile.Path`. Rejection cases walk `baseDir` via `filepath.WalkDir` and fail if any file landed outside `skillDir`, so this is a true traversal test — not just an error-message assertion.

Non-obvious calls:
- `InstallSkill`'s signature is unchanged; both existing callers (`runSingleInstall` and `installCatalogEntry` in `commands.go`) get the new validation for free.
- The absolute-path check runs on both `filePath` and `filepath.FromSlash(filePath)`. This is redundant on unix but necessary on Windows — see the inline comment.

Out-of-scope (flag for follow-up, not fixed here to keep the PR to one issue): `ParseIndex` in `index.go` filters only non-empty `Name`; a second layer of rejection there would produce nicer errors for `skill add --all` instead of the current per-entry `FAIL` with the raw wrapped error. Defensive duplication, not a functional gap.

## Verification

- [x] `make test` passes (`internal/skillmgr` 2.5s; all other packages green)
- [ ] Bug no longer reproduces: craft a local `index.json` with a malicious entry (e.g. `{"name":"../evil","files":["SKILL.md"]}` or `{"name":"pm","files":["../../etc/passwd"]}`), point `rawBaseURL` at a local httptest server, run `devpilot skill add pm`, and confirm the command errors out with a message naming the offending input and no file is written outside the expected skill directory.
- [x] No regressions in related functionality (existing `TestInstallSkill*` in `github_test.go` still pass)

## For Reviewers (human)

- [ ] Self-review of the code

Closes #74

### DIFF
--- a/internal/skillmgr/install.go
+++ b/internal/skillmgr/install.go
@@ -4,16 +4,30 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 // InstallSkill writes skill files into <baseDir>/<skillName>/.
 // baseDir is the absolute path to the skills directory (e.g., ".claude/skills" or "~/.claude/skills").
 // Existing files are silently overwritten.
+//
+// Both skillName and each SkillFile.Path can originate from the remote
+// skills/index.json, so the caller cannot assume they are trusted. Every
+// input is validated to prevent a malicious catalog from writing outside
+// skillDir via absolute paths or "../" segments.
 func InstallSkill(baseDir, skillName string, files []SkillFile) error {
+	if err := validateSkillName(skillName); err != nil {
+		return err
+	}
+
 	skillDir := filepath.Join(baseDir, skillName)
+	cleanSkillDir := filepath.Clean(skillDir)
 
 	for _, f := range files {
-		target := filepath.Join(skillDir, filepath.FromSlash(f.Path))
+		target, err := resolveSkillFilePath(cleanSkillDir, f.Path)
+		if err != nil {
+			return err
+		}
 		if err := os.MkdirAll(filepath.Dir(target), 0755); err != nil {
 			return fmt.Errorf("creating directory for %s: %w", f.Path, err)
 		}
@@ -22,4 +36,44 @@ func InstallSkill(baseDir, skillName string, files []SkillFile) error {
 		}
 	}
 	return nil
+}
+
+func validateSkillName(skillName string) error {
+	if skillName == "" {
+		return fmt.Errorf("invalid skillName %q: must not be empty", skillName)
+	}
+	if skillName == "." || skillName == ".." {
+		return fmt.Errorf("invalid skillName %q: must not be %q", skillName, skillName)
+	}
+	if strings.ContainsRune(skillName, '/') || strings.ContainsRune(skillName, '\\') ||
+		strings.ContainsRune(skillName, filepath.Separator) {
+		return fmt.Errorf("invalid skillName %q: must not contain path separators", skillName)
+	}
+	if filepath.IsAbs(skillName) {
+		return fmt.Errorf("invalid skillName %q: must not be an absolute path", skillName)
+	}
+	return nil
+}
+
+func resolveSkillFilePath(cleanSkillDir, filePath string) (string, error) {
+	if filePath == "" {
+		return "", fmt.Errorf("invalid file path %q: must not be empty", filePath)
+	}
+	// filepath.Join silently strips a leading separator, so "/etc/passwd"
+	// would land under skillDir as "etc/passwd". Reject absolutes first.
+	converted := filepath.FromSlash(filePath)
+	if filepath.IsAbs(filePath) || filepath.IsAbs(converted) || strings.HasPrefix(filePath, "/") {
+		return "", fmt.Errorf("invalid file path %q: must not be absolute", filePath)
+	}
+	target := filepath.Join(cleanSkillDir, converted)
+	cleaned := filepath.Clean(target)
+	// The "+ Separator" prefix check prevents "/a/skilldir-evil" from
+	// matching "/a/skilldir" as a descendant.
+	if cleaned != cleanSkillDir && !strings.HasPrefix(cleaned, cleanSkillDir+string(os.PathSeparator)) {
+		return "", fmt.Errorf("invalid file path %q: escapes skill directory", filePath)
+	}
+	if cleaned == cleanSkillDir {
+		return "", fmt.Errorf("invalid file path %q: resolves to skill directory itself", filePath)
+	}
+	return target, nil
 }
--- a/internal/skillmgr/install_test.go
+++ b/internal/skillmgr/install_test.go
@@ -1,0 +1,219 @@
+package skillmgr
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+// TestInstallSkillPaths covers path-traversal and input-validation behavior
+// of InstallSkill. Happy paths verify expected files land inside skillDir;
+// rejection paths verify a descriptive error is returned AND the filesystem
+// under baseDir contains no files outside skillDir.
+func TestInstallSkillPaths(t *testing.T) {
+	type wantFile struct {
+		relPath string // path under baseDir (forward slashes)
+		content string
+	}
+
+	cases := []struct {
+		name       string
+		skillName  string
+		files      []SkillFile
+		wantErr    bool
+		wantErrHas string // substring the error must contain (when wantErr)
+		wantFiles  []wantFile
+		skipOnWin  bool // skip on Windows when test depends on unix path semantics
+		onlyOnWin  bool
+	}{
+		{
+			name:      "happy path single file",
+			skillName: "pm",
+			files: []SkillFile{
+				{Path: "SKILL.md", Content: []byte("top")},
+			},
+			wantFiles: []wantFile{
+				{relPath: "pm/SKILL.md", content: "top"},
+			},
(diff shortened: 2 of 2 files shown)


SiyuQian (commented):
Hi @SiyuQian,

## PR Review: fix(skillmgr): reject traversal and separators in skill install paths (#86)

### TL;DR

Safe to merge for the traversal class the issue calls out: the new checks reject the listed attacks and the tests walk `baseDir` to prove nothing escaped. The one thing worth weighing before merge is whether to replace the hand-rolled containment with `os.Root` (Go 1.24+, available on this repo's Go 1.25.6) — it would also cover the symlink-escape path that the current check still allows.

### Unknown-Unknowns Sweep

1. Local pattern fit: matches existing `validateSkillAddArgs` style in `internal/skillmgr/commands.go` — `fmt.Errorf` validation strings, lowercase-start, no wrap on pure validation. Table-driven test style matches `TestParseIndex` in `index_test.go`.
2. Blast radius: `InstallSkill` signature unchanged. Two callers get validation for free (`runSingleInstall` and `installCatalogEntry` in `commands.go`). Minor user-visible behavior change: `devpilot skill add ./pm` used to resolve to `pm/` and succeed; it now errors on the `/` in the name. Probably acceptable but not mentioned in the PR description.
3. Known pitfalls (incl. security/data/reversibility): traversal + absolute-path + separator-in-name all covered for the current attack surface. Symlink-follow not covered — see Finding #2 below.
4. Stale-training check: `os.Root` / `os.OpenRoot` landed in Go 1.24 and covers this exact use case at the syscall level (on Linux via `openat2` with `RESOLVE_BENEATH`). `go.mod` is `go 1.25.6`, so it's usable. Worth a look before merging.
5. Hand-rolled vs. off-the-shelf: the containment logic is a clean hand-roll, but it duplicates what `os.Root` does natively — see Finding #1.

### Behavior Findings

#### Consider — Use `os.Root` instead of hand-rolled containment

- **Where:** `internal/skillmgr/install.go:18-39`, `:58-79`
- **Behavior today on this branch:** `InstallSkill` computes `filepath.Clean(target)` and asserts the cleaned path is under `cleanSkillDir + Separator`. That rejects `..` escapes and absolute paths correctly.
- **Why that's worth a look:** Go 1.24+ ships `os.Root` whose entire purpose is this check; it also enforces it at the syscall layer (Linux `openat2` with `RESOLVE_BENEATH`) so races and platform edge cases aren't a lift for us. The per-method doc is explicit: *"If any component of a file name passed to a method of Root references a location outside the root, the method returns an error. … Symbolic links must not be absolute."* Using it would replace ~40 lines of validation with `root, _ := os.OpenRoot(skillDir); root.WriteFile(...)` (plus wrapping `MkdirAll` via `root.Mkdir`). It also removes the need for the `HasPrefix(filePath, "/")` special case, the dual `FromSlash` + `IsAbs` check, and the `+ Separator` sibling-suffix guard — all of which are correct today but are hand-rolled versions of what `Root` does.
- **Suggested change:** Refactor `InstallSkill` to use `os.OpenRoot(skillDir)` and the `Root`-scoped I/O methods. Keep `validateSkillName` for the skillName (Root can't protect you from a bad directory name since it IS the directory name). Let the test that walks `baseDir` stay — it's still the right assertion.
- **Confidence:** medium — I've verified `os.Root` is in the stdlib at this Go version and read its doc; I have not prototyped the replacement to confirm every current test case still passes unchanged.

#### Consider — Pre-existing symlink inside skillDir can still be traversed

- **Where:** `internal/skillmgr/install.go:31-36`
- **Behavior today on this branch:** `os.MkdirAll` + `os.WriteFile` follow symlinks. If `skillDir/subdir` already exists as a symlink to `/etc`, and a catalog entry has `Path: "subdir/passwd"`, the cleaned target is `skillDir/subdir/passwd` which passes the containment check — but the syscall writes to `/etc/passwd`.
- **Why that's a problem:** A compromised catalog cannot *create* such a symlink via `InstallSkill` (it only writes regular files), so exploitability requires the symlink to already exist — e.g. a user who symlinked their skills directory for sync-across-machines, or a previous rogue skill installed by a different tool. Low likelihood, but the point of this PR is defense-in-depth against a compromised catalog, and this is a gap in that depth.
- **Suggested change:** `os.Root` fixes this too (per the doc quoted above: symlinks may not reference outside the root). If you keep the hand-rolled path, `os.Lstat` each directory component before `MkdirAll` and reject if any is a symlink, or `syscall.O_NOFOLLOW` on the write, or — simplest — `openat2(RESOLVE_BENEATH)` via `os.Root`.
- **Confidence:** low — requires a pre-existing symlink in the skill directory, which isn't something the catalog itself can plant.

### What's working well

- Tests don't just assert error strings — they walk `baseDir` with `filepath.WalkDir` on every rejection case and fail if any file landed outside `skillDir`. That's the right shape for a traversal test.
- Dual absolute-path check (`IsAbs(filePath) || IsAbs(converted) || HasPrefix(filePath, "/")`) is a correct defense-in-depth against Windows drive-letter vs. unix-style absolutes.
- Both non-obvious WHYs have one-line comments (`filepath.Join` silently stripping separators; the `+ Separator` suffix trick). Short, on-topic, not restating the code.

### Nits

- `internal/skillmgr/install.go:48-50` — `strings.ContainsRune(skillName, filepath.Separator)` is redundant with the explicit `'/'` and `'\\'` checks on every current Go platform. Can drop the third disjunct.
- `internal/skillmgr/install.go:45-46` — error body `"invalid skillName %q: must not be %q"` renders as `invalid skillName ".": must not be "."`, which reads as a self-referential tautology. `"must not be the current or parent directory"` (or similar) is clearer.
- `internal/skillmgr/install_test.go:30` — `onlyOnWin bool` is declared on the test struct but no case sets it. Dead field, drop it or add a case that uses it.

### Open Questions

- Does anyone legitimately invoke `devpilot skill add ./pm` or `devpilot skill add pm/` on the CLI today? If yes, the new rejection is a minor behavior break worth calling out in release notes; if no, silent is fine.

---

PR #87, opened 2026-04-25
fix(auth): non-blocking sends in OAuth callback handler

### LINKED ISSUE #77: [repo-scan:edge-case] OAuth callback handler can block on buffered resultCh if invoked twice

## What

`resultCh` is buffered with capacity 1. The handler can be invoked more than once (browser preflight/refresh, second request hitting the URL). The second send blocks forever because the caller only reads once before calling `srv.Shutdown`. Result: leaked goroutine and a potentially stuck Shutdown.

## Where

internal/auth/oauth.go — L130-L159

https://github.com/SiyuQian/devpilot/blob/3a48241fc258afc644ef6be9a9a30a94c1c629fb/internal/auth/oauth.go#L130-L159

## Evidence

```go
131      resultCh := make(chan callbackResult, 1)
132
133      mux := http.NewServeMux()
134      mux.HandleFunc("/callback", func(w http.ResponseWriter, r *http.Request) {
...
138              resultCh <- callbackResult{err: ErrAuthDenied}
...
145              resultCh <- callbackResult{err: errors.New("oauth: invalid state parameter (possible CSRF)")}
...
152              resultCh <- callbackResult{err: errors.New("oauth: missing authorization code in callback")}
...
158          resultCh <- callbackResult{code: code}
```

## Suggested fix

Send non-blocking (`select { case resultCh <- ...: default: }`) or use `sync.Once` to ensure only the first invocation writes.

## Metadata

- Category: `edge-case`
- Severity: `medium`
- Scanner: `edge-case-hunter`
- Scored: `75/100`

SiyuQian:
## Description

**Bug:** The `/callback` handler in `internal/auth/oauth.go:130` sent every result on a buffered channel of capacity 1 unconditionally. `StartFlow` reads exactly once from that channel and then runs `defer srv.Shutdown(context.Background())` (no deadline). A second request to `/callback` — browser refresh, preflight, a stale tab — would block its handler forever on the channel send, and `srv.Shutdown` would deadlock waiting for the in-flight handler to return. The blocked handler goroutine leaks.

**Cause:** `resultCh <- callbackResult{...}` blocks once the buffer is full. The handler had no guard for that.

**Fix:** Wrap every send site in `select { case resultCh <- ...: default: }`. The first invocation fills the buffer and is delivered to `StartFlow`; subsequent invocations short-circuit through the `default` branch and return normally. The HTTP response to the second request is still written normally — only the channel send is dropped.

## Review Guide

**Start here:** `internal/auth/oauth.go:130-159` — the four send sites in the callback handler are now wrapped in `select { ... default: }`. Channel capacity is unchanged. Return signature is unchanged.

Then `internal/auth/oauth_test.go:150-217` — `TestCallbackServerDoubleInvocation`. Table-driven across four second-call shapes (success, invalid state, denied, missing code). Each case:

1. Hits `/callback` with a valid success and drains `resultCh` (mirroring `StartFlow`'s read).
2. Hits `/callback` again with a 2-second `http.Client` timeout. Without the fix, the server would never write a response to the second request — the handler would be parked on the channel send.
3. Calls `srv.Shutdown` with a 2-second context. Without the fix, this would return `context.DeadlineExceeded` because the blocked handler keeps the server's request count above zero.

Both timeout-bounded assertions are tight enough to catch a regression: the unmodified code would fail at step 2 with the client timeout AND at step 3 with the context deadline.

Out-of-scope (flag for follow-up): the four send sites duplicate the `select-default` pattern. A small `trySend(ch, val)` helper would dedupe but is below the threshold to justify in a security-fix PR. Also: `StartFlow`'s `defer srv.Shutdown(context.Background())` uses an unbounded context — a `context.WithTimeout` there would be a second layer of protection in case some other future handler regresses.

## Verification

- [x] `make test` passes (`internal/auth` 0.5s; all packages green)
- [ ] Bug no longer reproduces: revert `internal/auth/oauth.go` to main, re-run `go test -run TestCallbackServerDoubleInvocation -timeout 10s ./internal/auth/`, observe the test hangs at the second `client.Get` and `srv.Shutdown` until the test timeout fires. With the fix, the test passes in ~0.1s.
- [x] No regressions in related functionality (existing `TestCallbackServer*` tests still pass)

## For Reviewers (human)

- [ ] Self-review of the code

Closes #77


SiyuQian (commented):
Hi @SiyuQian,

## PR Review: fix(auth): non-blocking sends in OAuth callback handler (#87)

### TL;DR

Safe to merge. The fix replaces every `ch <- ...` in the callback handler with the canonical `select { case ch <- ...: default: }`, the new test asserts both `client.Get` and `srv.Shutdown` complete within tight timeouts, and lint/test are green. One small Consider on test coverage and one Nit on duplicated boilerplate.

### Unknown-Unknowns Sweep

1. Local pattern fit: matches Go's standard non-blocking-send idiom; consistent with the existing `select { case <-resultCh: case <-time.After(...) }` in `StartFlow` so both ends of the channel use `select` now.
2. Blast radius: only `startCallbackServer` and one new test. The function's signature, channel capacity, and the `StartFlow` reader are all unchanged. No callers outside this file.
3. Known pitfalls (concurrency class):
   - Lost first result: not possible — the buffered channel is empty on first invocation, so the first send always succeeds via the case branch.
   - Race between simultaneous handlers: Go's select is atomic at runtime, so even with two concurrent handlers only one send completes; the other falls through to `default`. Safe.
   - srv.Shutdown semantics: with the fix, no handler is parked on a send, so Shutdown's "wait for in-flight requests" returns promptly. The unbounded `context.Background()` in `StartFlow`'s `defer srv.Shutdown(...)` (`internal/auth/oauth.go:350`) becomes harmless under this fix, but is still a latent risk if a future regression reintroduces a blocking handler. Worth a follow-up.
   - Reversibility / data: no state writes change.
4. Stale-training check: N/A — non-blocking send via `select-default` is a stable Go idiom that hasn't shifted.
5. Hand-rolled vs. off-the-shelf: `select-default` IS the off-the-shelf primitive. The four duplicated send sites could be folded behind a tiny helper; see Nit below.

### Behavior Findings

#### Consider — Test does not cover simultaneous double invocation

- **Where:** `internal/auth/oauth_test.go:152-217`
- **Behavior today on this branch:** The test issues two sequential `client.Get` calls. The first completes in full (including draining `resultCh`) before the second starts.
- **Why that's worth a look:** The race the fix actually closes is two handlers that *both observe an empty channel* before either send completes — i.e., near-simultaneous arrivals. Sequential calls exercise the post-buffer-full path but not the concurrent one. Go's runtime makes simultaneous select-with-default safe, so there's no real bug here, but a test that fires two goroutines at the same listener with a `sync.WaitGroup` would (a) prove the concurrent path, and (b) fail visibly if a future refactor reintroduces unguarded sends. With the current sequential test, a regression that only manifests under contention would slip through.
- **Suggested change:** Add one more table case (or a sibling test) that spawns two concurrent `client.Get` calls and asserts both return within the client timeout AND `srv.Shutdown` returns within its context.
- **Confidence:** medium — the assertion is correct that the current test doesn't cover concurrency; it's a judgment call whether the additional coverage is worth the extra ~20 lines.

### What's working well

- The test asserts both `client.Get` AND `srv.Shutdown` complete within tight 2-second timeouts. That's the right pair: the unmodified code fails on both ends, so this is a regression-tight assertion.
- Four table cases (success / invalid state / denied / missing code) cover every send site individually, so a future change that forgets to wrap one of the four is caught.
- One concise comment on the handler explaining the non-obvious WHY ("only the first result wins"). No narration on the obvious select-default mechanics. Matches CLAUDE.md's comment policy.

### Nits

- `internal/auth/oauth.go:134-171` — the `select { case ch <- ...: default: }` block is repeated four times. A two-line helper `func trySend[T any](ch chan<- T, v T) { select { case ch <- v: default: } }` would dedupe to four call sites of `trySend(resultCh, ...)`. Optional; the verbatim pattern is unambiguous as-is.
- `internal/auth/oauth.go:134` — the explanatory comment is on the `mux.HandleFunc(...)` line, not on the handler body. Reads slightly oddly; the comment applies to the handler's send sites, not to mux registration. Moving it inside the `func(w, r)` (one line above the first `if errParam`) would localize the WHY.

### Open Questions

- Does anything else in the auth flow rely on `srv.Shutdown` blocking until a *specific* handler returns (e.g. to ensure logs flush)? Spot-check says no — the handler only writes HTML and a channel value. If yes, please flag.

---

PR #88, opened 2026-04-25
fix(generate): fall back to single commit when validation empties plan

### LINKED ISSUE #78: [repo-scan:edge-case] executeCommit panics if validatePlan empties commit list

SiyuQian:
## Description

**Bug:** `validatePlan` in `internal/generate/commit_plan.go` could return a `commitPlan` with `len(Commits) == 0`. `commitModel.startExecution` (`internal/generate/commit_model.go:233-243`) then called `m.executeCommit(0)`, which evaluates `m.plan.Commits[0]` at `commit_model.go:307` and panicked the TUI.

**Cause:** Reachable scenario in `validatePlan`:

1. Each plan commit's `Files` is filtered to only-staged files (`commit_plan.go:77-88`). If every file in a commit is unknown, that commit's `Files` becomes `nil`.
2. Uncovered staged files are appended to the *last* commit (`:91-102`). If every staged file is in `Excluded`, no missing files exist, so this step adds nothing.
3. The final filter at `:104-111` drops zero-file commits. If every commit was emptied in step 1, the result is `[]`.

The author of `commit_model.executeCommit` was reasonably entitled to assume `Commits` was non-empty — the contract violation is at `validatePlan`'s exit, not at the consumer.

**Fix:** After the existing nonEmpty filter, if `plan.Commits` is empty, repopulate it via the existing `fallbackPlan(\"\", stagedFiles)` helper (which already handles the \"give up and use a single commit\" pattern in `parseCommitPlan`). Append a warning so the user sees the fallback was triggered.

Deliberately *not* changing `executeCommit`. The bug was the contract violation in `validatePlan`; once the contract is restored, defending against an impossible scenario in the consumer would be exactly the kind of \"error handling for scenarios that can't happen\" CLAUDE.md asks against.

## Review Guide

**Start here:** `internal/generate/commit_plan.go:113-122` — the new fallback block at the end of `validatePlan`. Reuses `fallbackPlan` rather than duplicating its body. Single warning message.

Then `internal/generate/commit_plan_test.go` — `TestValidatePlan_FallbackOnEmpty`, table-driven with two cases that both reduce to empty after validation:
- One commit referencing only unknown files, only staged file in `Excluded`.
- Multi-commit variant where all referenced files are unknown and all staged files are excluded.

Both assert exactly one fallback commit, the expected files, and the fallback warning.

Out-of-scope (flagged for follow-up): `commit_plan.go:99` does `last := len(plan.Commits) - 1` without a guard. With the current `parseCommitPlan` always returning ≥1 commit on entry, this is fine, but if `validatePlan` ever became reachable from a different caller with an empty plan input, it would index `[-1]`. A two-line guard would harden it; not in scope here.

## Verification

- [x] `make test` passes (`internal/generate` 0.2s; all packages green)
- [ ] Bug no longer reproduces: revert the new block in `validatePlan`, run `go test -run TestValidatePlan_FallbackOnEmpty -timeout 5s ./internal/generate/`. Without the fix the test panics on `validated.Commits[0]` access; with the fix it passes in <1s.
- [x] No regressions: existing `TestValidatePlan_UnknownFile` and `TestValidatePlan_MissingFile` still pass.

## For Reviewers (human)

- [ ] Self-review of the code

Closes #78

### DIFF
--- a/internal/generate/commit_plan.go
+++ b/internal/generate/commit_plan.go
@@ -110,6 +110,17 @@ func validatePlan(plan commitPlan, stagedFiles []string) (commitPlan, []string) 
 	}
 	plan.Commits = nonEmpty
 
+	// If validation removed every commit, fall back to a single commit so
+	// downstream consumers (e.g. commitModel.executeCommit) keep their
+	// "plan.Commits is non-empty" invariant. This can happen when every
+	// commit references only unknown files and every staged file is
+	// already in Excluded.
+	if len(plan.Commits) == 0 {
+		fallback := fallbackPlan("", stagedFiles)
+		plan.Commits = fallback.Commits
+		warnings = append(warnings, "plan validation removed all commits; falling back to a single commit")
+	}
+
 	return plan, warnings
 }
 
--- a/internal/generate/commit_plan_test.go
+++ b/internal/generate/commit_plan_test.go
@@ -108,6 +108,79 @@ func TestValidatePlan_MissingFile(t *testing.T) {
 	}
 }
 
+func TestValidatePlan_FallbackOnEmpty(t *testing.T) {
+	cases := []struct {
+		name        string
+		plan        commitPlan
+		staged      []string
+		wantFiles   []string
+		wantMessage string
+	}{
+		{
+			name: "all commit files unknown and only staged file excluded",
+			plan: commitPlan{
+				Commits: []commitEntry{
+					{Message: "feat: ghost", Files: []string{"ghost.go"}},
+				},
+				Excluded: []excludedFile{
+					{File: "real.go", Reason: "Contains secrets"},
+				},
+			},
+			staged:      []string{"real.go"},
+			wantFiles:   []string{"real.go"},
+			wantMessage: "chore: update files",
+		},
+		{
+			name: "multiple commits all reference unknown files and all staged files excluded",
+			plan: commitPlan{
+				Commits: []commitEntry{
+					{Message: "feat: a", Files: []string{"ghost1.go"}},
+					{Message: "feat: b", Files: []string{"ghost2.go"}},
+				},
+				Excluded: []excludedFile{
+					{File: "x.go", Reason: "secret"},
+					{File: "y.go", Reason: "secret"},
+				},
+			},
+			staged:      []string{"x.go", "y.go"},
+			wantFiles:   []string{"x.go", "y.go"},
+			wantMessage: "chore: update files",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			validated, warnings := validatePlan(tc.plan, tc.staged)
+
+			if len(validated.Commits) != 1 {
+				t.Fatalf("expected exactly 1 fallback commit, got %d", len(validated.Commits))
+			}
+			if validated.Commits[0].Message != tc.wantMessage {
+				t.Errorf("expected fallback message %q, got %q", tc.wantMessage, validated.Commits[0].Message)
+			}
+			if len(validated.Commits[0].Files) != len(tc.wantFiles) {
+				t.Fatalf("expected %d files in fallback commit, got %d", len(tc.wantFiles), len(validated.Commits[0].Files))
+			}
+			for i, f := range tc.wantFiles {
+				if validated.Commits[0].Files[i] != f {
+					t.Errorf("file[%d]: expected %q, got %q", i, f, validated.Commits[0].Files[i])
+				}
+			}
+
+			foundFallbackWarning := false
+			for _, w := range warnings {
+				if strings.Contains(w, "falling back") || strings.Contains(w, "fallback") {
+					foundFallbackWarning = true
+					break
+				}
+			}
+			if !foundFallbackWarning {
+				t.Errorf("expected a warning mentioning fallback, got %v", warnings)
+			}
+		})
+	}
+}
+
 func TestTruncateDiff_SmallDiff(t *testing.T) {
 	diff := "diff --git a/file.go b/file.go\n--- a/file.go\n+++ b/file.go\n@@ -1 +1 @@\n-old\n+new\n"
 	result := truncateDiff(diff)


SiyuQian (commented):
Hi @SiyuQian,

## PR Review: fix(generate): fall back to single commit when validation empties plan (#88)

### TL;DR

Holding for a Blocking concern: the fallback path triggers exactly when every staged file is in `Excluded`, and it then puts those same excluded files into the fallback commit's `Files`. Combined with `executeCommit`'s `git reset` + `git add -- <files>` sequence, that re-stages and commits files the user explicitly excluded — including, plausibly, secrets. The PR fixes the panic but trades it for a worse failure mode on the same trigger.

### Unknown-Unknowns Sweep

1. Local pattern fit: reuses the existing `fallbackPlan` helper rather than duplicating it. Matches the style in `parseCommitPlan` (line 36). Good.
2. Blast radius: only `validatePlan`'s caller (`commit_model.analyzeChanges`) is affected, but the resulting `plan.Commits[0].Files` flows into `executeCommit` which calls `git add -- <files>` after `git reset HEAD -- .` — so the contents of `Files` end up in a real commit. See Finding #1.
3. Known pitfalls (data integrity / reversibility): the trigger condition for this fallback is narrower than the issue described. I traced it: with non-empty staged files, the `missing` block at `commit_plan.go:91-102` re-fills the last commit. The result is empty *only* when every staged file is in `Excluded` (so `missing == []`) AND every plan commit's files are filtered out. That intersection means the fallback runs precisely when the user has explicitly opted out of every staged file. See Finding #1.
4. Stale-training check: N/A — straight Go.
5. Hand-rolled vs. off-the-shelf: reuses `fallbackPlan`. No duplication.

### Behavior Findings

#### Blocking — Fallback re-commits files the user just excluded

- **Where:** `internal/generate/commit_plan.go:118-122` interacting with `internal/generate/commit_model.go:233-243` and `:306-332`.
- **Behavior today on this branch:** When `validatePlan` empties `Commits`, the new block calls `fallbackPlan("", stagedFiles)`, producing one commit whose `Files` is the *full* `stagedFiles`. The trigger condition (traced through `validatePlan`'s logic) requires every staged file to be in `Excluded` — otherwise the missing-files block at `commit_plan.go:91-102` would have re-populated the last commit and the empty filter wouldn't have stripped it. So at the moment of fallback, `stagedFiles ⊆ {files in plan.Excluded}`.

  `startExecution` then runs `tea.Sequence(m.unstageExcluded(), m.executeCommit(0))`. `unstageExcluded` does `git reset HEAD -- <excluded>` (best-effort, errors ignored), but `executeCommit` immediately runs `git reset HEAD -- .` (unstaging everything), then `git add -- <commit.Files...>` (which is the same set the user excluded), then `git commit`. The exclusion is silently undone.

  Concrete repro: user stages `.env`, `secrets.yml`, marks both as Excluded in the editor. Claude (legitimately confused by an exclusion-only plan, or hallucinating) returns a plan whose only commit references some other file. Validation strips the unknown commit; `missing == []` because both staged files are excluded; commits is empty; fallback engages; commit gets `Files: [".env", "secrets.yml"]`; commit lands.
- **Why that's a problem:** This is a security regression vs. main. On main the user gets a panic and no commit. With this PR they get a commit that contains exactly the files they told the tool to exclude — likely secrets, given that the test fixture itself uses `Reason: "Contains secrets"` for the excluded entry. The single warning string `"plan validation removed all commits; falling back to a single commit"` does not mention exclusions and is unlikely to be parsed as "we just committed everything you told us not to."
- **Suggested change:** Choose one of the following before merging:
   1. **Build the fallback from `staged \ excluded`.** If the result is non-empty, commit those files. If the result IS empty (the actual no-op case), refuse and surface an error the TUI can render — `validatePlan` would need to grow an error return, or use a sentinel `commitPlan{}` that the caller checks. This preserves the user's exclusions.
   2. **Don't fall back; surface an error.** Change `validatePlan` to return `(commitPlan, []string, error)` and have `analyzeChanges` route the error into the existing error UI in the TUI. Removes the panic without introducing a security regression. More invasive but unambiguous.
   3. **Keep the current fallback but also add a defensive bounds check in `executeCommit`** so the user gets a friendly error instead of either a panic or a sneaky commit. Weaker than 1 or 2 but at least non-destructive.

  The original issue's suggested fix included "additionally guard `executeCommit` with `if index >= len(m.plan.Commits) {...}`" — that aligns with option 3 and would have avoided this trade-off.
- **Confidence:** medium-high — I traced both `validatePlan` (to confirm the trigger requires `stagedFiles ⊆ Excluded`) and `executeCommit` (`git reset .` + `git add -- <files>` undoes the prior unstage). I have not run a live repro of the secret-leak scenario; the inference is from the code paths and the new test's own fixture (which uses `Reason: "Contains secrets"`).

#### Should-fix — Test asserts the regressing behavior is correct

- **Where:** `internal/generate/commit_plan_test.go:130-138` and `:150`
- **Behavior today on this branch:** Both new test cases assert `wantFiles == staged` — i.e., the test *codifies* that a previously-excluded file lands in the fallback commit.
- **Why that's a problem:** Once Finding #1 is addressed, the assertion will need to flip — the fallback should NOT contain the excluded files (or there should be no fallback at all). Leaving the test as-is locks in the regression. If you take option 1 above, `wantFiles` becomes `staged \ excluded` (which is `[]` for both current cases, so the test would need a different shape — likely "expect an error / sentinel"). If option 2, the test asserts the new error.
- **Suggested change:** Restructure the test alongside the fix in #1.
- **Confidence:** high — direct read of the assertions.

### What's working well

- Reuses `fallbackPlan` instead of duplicating the helper. Single source of truth for the "give up gracefully" message.
- Comment block above the new `if len(plan.Commits) == 0` correctly explains *when* the path is reached. (Whether it should be reached at all is the Blocking finding above.)
- PR description is honest about scope: it explicitly chose not to guard `executeCommit`, citing CLAUDE.md's "no error handling for scenarios that can't happen." That tradeoff is reasonable in isolation; the Blocking finding shows that the scenario is reachable and is now silent rather than loud.

### Nits

- `internal/generate/commit_plan.go:99` — `last := len(plan.Commits) - 1` has no guard for empty input. Out of scope per the PR's stated guard-rails, but called out in the description as a follow-up — fine.

### Open Questions

- The PR description observes "the author of `executeCommit` was reasonably entitled to assume `Commits` was non-empty." Agreed, but `validatePlan`'s contract was always "may produce an empty plan" — both the comment "Returns warnings and a corrected plan" and the inline filter at `:104-111` advertise that. Was the design intent that callers check `len(plan.Commits)` before iterating? If so, the panic was a missing check at the consumer; the fix here moves the contract closer to "always non-empty" but as a side-effect changes commit content, which is the more dangerous failure mode.

---

PR #89, opened 2026-04-25
fix(initcmd): match gitignore entries by line, not substring

### LINKED ISSUE #79: [repo-scan:edge-case] EnsureGitignore uses substring match — misses real entries

## What

`strings.Contains(existing, entry)` is a substring match, not a line match. An entry like `.devpilot` is treated as present if the gitignore already contains `.devpilot.yaml` or `!.devpilot-keep`. The intended ignore is silently skipped — e.g. a logs directory under `.devpilot/` is never ignored.

## Where

internal/initcmd/generate.go — L142-L147

https://github.com/SiyuQian/devpilot/blob/3a48241fc258afc644ef6be9a9a30a94c1c629fb/internal/initcmd/generate.go#L142-L147

## Evidence

```go
142      var toAdd []string
143      for _, entry := range entries {
144          if !strings.Contains(existing, entry) {
145              toAdd = append(toAdd, entry)
146          }
147      }
```

## Suggested fix

Split `existing` on newlines, strip whitespace and leading `!` on each line, then compare each line exactly against `entry`.

## Metadata

- Category: `edge-case`
- Severity: `low`
- Scanner: `edge-case-hunter`
- Scored: `75/100`

SiyuQian:
## Description

**Bug:** `EnsureGitignore` (`internal/initcmd/generate.go:144`) checked for already-present entries with `strings.Contains(existing, entry)` — a substring match, not a line match. The intended ignore could be silently skipped if it appeared as a substring of any other line:

- existing line `.devpilot/logs/extra` → requested `.devpilot/logs/` → falsely matched, not added.
- existing comment `# .devpilot/logs/` → falsely matched, entry not added.
- existing negation `!.devpilot/logs/` → falsely matched, but should still be treated as "present" — see Cause.

**Cause:** Substring vs. line semantics. The git scanner's example (`.devpilot.yaml` matching `.devpilot`) is the same shape: entry `.devpilot` would silently match `.devpilot.yaml`.

**Fix:** New unexported helper `gitignoreContains(existing, entry string) bool` walks `existing` line by line, and for each line:

1. trims surrounding whitespace,
2. skips blank and `#`-comment lines,
3. strips a single leading `!` (gitignore negation; `!foo` and `foo` are the same identifier for the "is this declared?" check),
4. compares to the trimmed, leading-`!`-stripped requested `entry` with `==`.

The block-write logic, `\n# DevPilot\n` prefix, and trailing-newline guard are unchanged.

## Review Guide

**Start here:** `internal/initcmd/generate.go:133-152` — the new `gitignoreContains` helper. Single call site at `:165`. Replaces one `strings.Contains` call; no other behavior changes.

Then `internal/initcmd/generate_test.go` — `TestEnsureGitignore`, table-driven with seven subtests:

- `adds entry to missing file` — no `.gitignore` exists; entry appended.
- `adds entry when only substring matches exist` — gitignore contains `.devpilot/logs/extra`; entry `.devpilot/logs/` IS added (regression case).
- `skips entry when exact line match exists` — entry already present.
- `skips entry when negated line match exists` — `!entry` already present.
- `treats lines with surrounding whitespace as equal` — `  entry  ` already present.
- `ignores comment lines` — `# entry` does not satisfy the check; entry IS added.
- `multiple entries, partial overlap` — only the missing one is appended.

Plus a separate `TestEnsureGitignore_DoesNotDuplicateExactMatch` and a small file-level `gitignoreHasLine` helper used by the assertions.

Out-of-scope (flag for follow-up): the existing `block += entry + "\n"` loop at `generate.go:186-189` is a `string += string in a loop` (the IDE flags it). Pre-existing; will rewrite to `strings.Builder` in a separate cleanup PR if desired.

## Verification

- [x] `make test` passes (`internal/initcmd` 0.2s; all packages green)
- [ ] Bug no longer reproduces: revert the `gitignoreContains` call back to `strings.Contains(existing, entry)` and re-run `go test -run TestEnsureGitignore -timeout 5s ./internal/initcmd/`. The "only substring matches exist" and "ignores comment lines" subtests fail; with the fix they pass.
- [x] No regressions: existing `internal/initcmd/` tests still pass.

## For Reviewers (human)

- [ ] Self-review of the code

Closes #79


SiyuQian (commented):
Hi @SiyuQian,

## PR Review: fix(initcmd): match gitignore entries by line, not substring (#89)

### TL;DR

Safe to merge. The fix replaces the substring check with a line-equal helper that also handles negation, whitespace, and comments. Coverage is solid. One Consider on test/impl helper duplication and one Nit on the existing `string += string` loop pre-dating this PR.

### Unknown-Unknowns Sweep

1. Local pattern fit: matches `internal/initcmd` test style (`t.TempDir`, `os.WriteFile` fixtures, table-driven subtests). Helper extraction style is consistent with other small unexported helpers in the package.
2. Blast radius: single caller (`internal/initcmd/commands.go:101` passing `.devpilot/logs/`). Behavior change is strictly recovering missed adds — there's no scenario where the old `strings.Contains` returned `false` and the new helper returns `true`, so no regression on previously-correct behavior. (The opposite — `Contains==true`, `gitignoreContains==false` — IS the case the issue reports, and the fix is to add an entry that previously got silently dropped.)
3. Known pitfalls (input boundary):
   - Glob patterns (`*.log`) — line-equal works because we're comparing pattern strings, not matched files.
   - Trailing-slash directory-only patterns (`foo/` vs `foo`) — preserved by `==`.
   - Escaped leading `\#` for paths starting with `#` — `HasPrefix(line, "#")` is false for `\#foo`, so it's correctly treated as a real entry. Verified by trace.
   - Negation semantics: the helper treats `!foo` and `foo` as the same identifier for "is this declared?" — opinionated but correct for the intent ("don't re-declare"). See Open Questions for the case where this might surprise.
4. Stale-training check: N/A — straight `strings` use.
5. Hand-rolled vs. off-the-shelf: gitignore libraries exist (e.g. `go-gitignore`) but pulling a dep for a 15-line line-equality check is overkill. Hand-rolled is the right call.

### Behavior Findings

#### Consider — Test helper is a near-clone of the production helper

- **Where:** `internal/initcmd/generate_test.go:166-180` (`gitignoreHasLine`) duplicates the body of `internal/initcmd/generate.go:138-152` (`gitignoreContains`).
- **Behavior today on this branch:** Both helpers walk the same lines, apply the same trims, and compare with `==`. They differ only in name and visibility. The test helper exists specifically so the assertions don't call back into the function under test — fair, but the result is two near-identical implementations that can drift.
- **Why that's a problem:** If the production helper has a subtle parser bug (say, mishandling `\r\n` line endings, or escape semantics), the test helper has the same bug, and the test suite will pass while the user-visible behavior is broken. The substring regression caught here would have been caught either way, but a future bug specifically in the line-extraction step could escape both.
- **Suggested change:** Either (a) drop `gitignoreHasLine` and have at least one subtest read the raw file and assert the literal content (e.g. for "adds entry when only substring matches exist", assert `strings.HasSuffix(got, ".devpilot/logs/\n")`), giving you a parser-independent ground-truth check; or (b) keep `gitignoreHasLine` for ergonomic reuse but add one such literal-content assertion in the regression case as a tripwire. The other subtests can keep using the helper.
- **Confidence:** medium — the duplication is visible in the diff; the practical impact is low because the helper is short and easily code-reviewed.

### What's working well

- Tight scope: one helper extracted, one `strings.Contains` swapped, no other behavior changed. The block-writer, `\n# DevPilot\n` prefix, and trailing-newline guard are untouched.
- Test cases cover every meaningful shape: missing file, empty file, substring-only (regression), exact match, negated match, surrounding whitespace, comment line, multi-entry partial overlap. The "ignores comment lines" case is the under-appreciated one — it confirms `# .devpilot/logs/` does NOT count as a declaration.
- The `TestEnsureGitignore_DoesNotDuplicateExactMatch` separate test asserts the file bytes are unchanged (`string(data) != initial`), which is a stronger guarantee than just "entry is present" — it catches accidental rewrites/normalization.
- Helper has a single comment block explaining trim, negation, and skip rules. Stays under five lines, no narration of mechanics.

### Nits

- `internal/initcmd/generate.go:188` — `block += entry + "\n"` inside a loop. Pre-existing; flagged as `stringsbuilder` by the IDE but not by `make lint`. Out of scope for this PR (correct call to keep it out per the description). Worth a separate two-line refactor to `strings.Builder` if the file is touched again.
- `internal/initcmd/generate.go:141`, `generate_test.go:171` — `strings.Split(existing, "\n")` allocates the slice; Go 1.24+ has `strings.SplitSeq` for iterator-based traversal. IDE flags it; `make lint` doesn't. Same trade-off — consistency with other `strings.Split` calls in the repo wins.

### Open Questions

- Negation: if a user's `.gitignore` legitimately contains `!.devpilot/logs/` (they want this directory NOT to be ignored, perhaps overriding a parent-level `.devpilot/`), the new behavior treats that as "already declared" and skips the requested `.devpilot/logs/` add. That's almost certainly the right call — re-declaring it positively would just make a confusing back-and-forth — but worth a sanity check from someone who's actually used negation in this project.

---

PR #108, opened 2026-05-19
feat(graph): Phase 1 — Go parser + SQLite store + cross-file resolver

SiyuQian:
## Summary

First phase of the **devpilot graph** subsystem — a Go-native code graph built into the devpilot CLI for token-efficient PR review and structural code analysis.

This PR ships the **Phase 1 foundation** per [the implementation plan](./docs/plans/2026-05-19-devpilot-graph-plan.md):

- SQLite-backed graph store using pure-Go `modernc.org/sqlite` (preserves single-binary cross-compilation)
- Tree-sitter Go parser extracting file / function / method / struct / interface / type nodes
- Edge extraction: `contains` / `calls` / `imports` / `tests` / `implements`
- Cross-file resolver: rewrites synthetic `external::<Name>` references to intra-module node IDs and infers Go interface implementations by exact method-set match

12 tasks, 12 tests, ~2700 LOC across `internal/graph/{parser,store,resolver}`. No CLI surface yet — that lands in Phase 4.

## Scope: Phase 1 only

The full plan covers seven phases (~13 weeks total). This PR ships Phase 1 as a reviewable unit per the plan's "Phase boundary = a tag" guidance (`graph-phase-1` tagged at HEAD).

- **In this PR:** Go parser, SQLite store, cross-file/interface resolver
- **Future PRs:** TS/JS/Rust parsers + cache (Phase 2), query layer (Phase 3), CLI subcommands (Phase 4), LSP cross-check tests (Phase 5), skill integration (Phase 6), polish (Phase 7)

Each subsequent phase re-invokes `superpowers:writing-plans` against itself to expand half-day tasks into bite-sized TDD steps with the latest context.

## Design references

- [Design spec](./docs/plans/2026-05-19-devpilot-graph-design.md) — full architecture, rationale, decision log
- [Implementation plan](./docs/plans/2026-05-19-devpilot-graph-plan.md) — phased task breakdown

## Test plan

- [x] `go test ./internal/graph/... -count=1` — 12 tests pass across 3 packages
- [x] `go build ./...` clean
- [ ] Fresh clone, `go mod download`, `go test ./...` passes
- [ ] Spot-check parser output against `gopls` for one fixture (manual)
- [ ] Confirm `Greeter.Hello` is correctly extracted as `method` with `Container=Greeter`
- [ ] Confirm `Console implements Greeter` is emitted while `Mute implements Greeter` is NOT


internal/graph/parser/go.go
```diff
@@ -0,0 +1,345 @@
...
+// NewGoParser returns a Parser for Go source files.
+func NewGoParser() *GoParser { return &GoParser{lang: goLang.GetLanguage()} }
+
+func (p *GoParser) Language() string     { return "go" }
+func (p *GoParser) Extensions() []string { return []string{".go"} }
+
+// Parse extracts the file node and top-level function declarations.
+// Methods, types, calls, imports, and tests edges are added in later tasks.
+func (p *GoParser) Parse(path string, src []byte) (ParseResult, error) {
+	parser := sitter.NewParser()
```
SiyuQian, line 27: **Should-fix · Confidence 95**

`sitter.NewParser()` is created on every `Parse()` call but never closed. The underlying `ts_parser*` is C-heap memory that Go's GC does not see; in a full repo build (thousands of files) this leaks one C parser handle per file.

The sibling `defer tree.Close()` on the result handles the tree, but the parser itself is separate. `*sitter.Parser` has its own `Close()` method.

Fix:
```go
parser := sitter.NewParser()
defer parser.Close()
parser.SetLanguage(p.lang)
```

Alternative: move `*sitter.Parser` into `GoParser` as a reusable field, but tree-sitter parsers are not goroutine-safe so you'd need a `sync.Mutex` — and that breaks the concurrency claim in `parser.go:9`. Per-call allocation + defer is the simplest fix that preserves the stated contract.

internal/graph/parser/go.go
```diff
@@ -0,0 +1,345 @@
...
+				IsExported: isExportedGo(name),
+			})
+			res.Edges = append(res.Edges, store.Edge{Src: path, Dst: id, Kind: "contains"})
+			if bodyNode := child.ChildByFieldName("body"); bodyNode != nil {
+				callEdges := walkCalls(bodyNode, src, path, id, intraFileFuncs)
+				res.Edges = append(res.Edges, callEdges...)
+
+				if isGoTestFunc(name, child, src) {
+					for _, e := range callEdges {
+						res.Edges = append(res.Edges, store.Edge{Src: e.Src, Dst: e.Dst, Kind: "tests"})
```
SiyuQian, line 84: **Consider · Confidence 85**

`tests` edges are emitted to **every** callee of a `TestXxx` body — including `t.Fatal`, `t.Helper`, `fmt.Println`, test helpers, and the actual function under test. After full repo build, every test will appear to "test" `external::t.Fatal`, etc.

Design spec §3 says `tests` edges represent "this test covers this function". The current heuristic over-approximates by ~10x in real repos.

Minimum fix: skip `external::` destinations.
```go
if isGoTestFunc(name, child, src) {
    for _, e := range callEdges {
        if strings.HasPrefix(e.Dst, "external::") {
            continue
        }
        res.Edges = append(res.Edges, store.Edge{Src: e.Src, Dst: e.Dst, Kind: "tests"})
    }
}
```

More precise alternative (later phase): match `TestFoo` → intra-file `Foo` by naming convention.

internal/graph/parser/go.go
```diff
@@ -0,0 +1,345 @@
...
+	case "interface_type":
+		return "interface"
+	default:
+		return "type"
+	}
+}
+
+// isGoTestFunc returns true if name starts with "Test" and the function has
+// a parameter of type *testing.T.
+func isGoTestFunc(name string, fn *sitter.Node, src []byte) bool {
```
SiyuQian, line 293: **Consider · Confidence 90**

Go's `go test` requires that the character after `Test` is **uppercase** (or a digit) — `func Testfoo(t *testing.T)` is NOT executed by `go test`. The current check only validates the 4-char prefix, so it will produce spurious `tests` edges for non-test helpers like `func TestableState(...)` or `func Testfoo(...)`.

Fix:
```go
if !strings.HasPrefix(name, "Test") {
    return false
}
if len(name) > 4 && unicode.IsLower([]rune(name)[4]) {
    return false
}
```

(`unicode` is already imported.)

internal/graph/parser/go.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// importSpecPath extracts the import path from an import_spec, stripping the
+// quotes. For aliased imports (e.g. `alias "strings"`) it returns the path,
+// not the alias. Returns the final path segment (e.g. "strings" for
+// "github.com/foo/strings") to keep external IDs short.
+//
+// For v1 we keep just the last segment to match the test expectation. Later
+// tasks can refine this to keep the full path when needed.
+func importSpecPath(spec *sitter.Node, src []byte) string {
```
SiyuQian, line 335: **Consider · Confidence 88**

`importSpecPath` truncates to the last path segment. `"fmt"`, `"github.com/foo/fmt"`, and `"go.example.com/fmt"` all map to the same `external::fmt` ID. In a monorepo this means `imports` edges silently merge unrelated packages.

This is fine for v1 if you're confident no real consumer relies on `external::` IDs being unique across modules — but the SQLite store treats `(src, dst, kind)` as the edge primary key, so colliding IDs deduplicate edges that should be distinct. Fixing this post-v1 will require a schema migration of the `dst` column.

Suggested fix (one-line change, no schema impact): store the **full** path as the external ID:
```go
return raw // skip the lastSlash() truncation; ID becomes external::github.com/foo/bar
```

The last-segment display name can be derived at query time when needed for human-readable output. Drop `lastSlash` if it ends up unused.

internal/graph/resolver/implements.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	// 2. Populate method sets: walk method nodes and match to their struct.
+	for _, r := range results {
+		for _, n := range r.Nodes {
+			if n.Kind == "method" && n.Container != "" {
+				// Match to a struct whose ID ends with "::<Container>".
+				suffix := "::" + n.Container
+				for _, si := range structs {
+					if endsWith(si.id, suffix) {
```
SiyuQian: **Should-fix · Confidence 78**

`endsWith(si.id, "::"+n.Container)` attributes a method to every struct sharing the same simple name across files. If `a.go` and `b.go` both define `type Buffer struct{}`, and `c.go` has `func (Buffer) Write()` (a Go syntax error, but the parser doesn't check), or if a method's `Container` field is just the bare receiver type name, the suffix match will hit both `a.go::Buffer` and `b.go::Buffer`. Both could then be incorrectly declared to implement an interface.

The resolver's name-index for calls (resolver.go:21) documents the "first-occurrence wins" limitation. The same ambiguity exists unacknowledged for struct method attribution.

Fix: in Go, methods must be declared in the same package as their receiver — which in practice means the same `file_*.go` file group within a package. Compare full qualified ID, not just suffix:
```go
for _, si := range structs {
    // Methods must share the same file as their struct (within a Go package).
    // Comparing the path portion of si.id with n.Path eliminates cross-file false matches.
    if si.id == n.Path+"::"+n.Container {
        si.methods[n.Name] = struct{}{}
        break
    }
}
```
(Methods can technically be in any file of the same package, not just the file defining the struct — so if you ever add multi-file-per-package fixtures, refine this to package-scoped instead of file-scoped.)

internal/graph/parser/parser.go
```diff
@@ -0,0 +1,28 @@
+// Package parser defines the common interface for language-specific
+// tree-sitter parsers that produce nodes and edges for the code graph.
+package parser
+
+import "github.com/siyuqian/devpilot/internal/graph/store"
+
+// Parser parses source files of a single language into ParseResult.
+// Implementations are stateless and safe for concurrent use.
+type Parser interface {
```
SiyuQian, line 14: **Nit · Confidence 85**

`// Implementations are stateless and safe for concurrent use.` is currently honored only because each `Parse()` allocates a fresh `*sitter.Parser`. If you fix the leak in `go.go:26` by hoisting a shared parser onto `GoParser` (the obvious alternative), this comment becomes a lie and concurrent calls will race on the C state.

If the planned fix is per-call allocation + `defer parser.Close()`: keep this comment but expand it to make the contract explicit:
```go
// Parser parses source files of a single language into ParseResult.
// Implementations must not hold mutable state across Parse calls so they can
// be safely invoked from multiple goroutines.
type Parser interface { ... }
```
This way a future maintainer can't quietly add a shared *sitter.Parser cache without noticing the contract.

SiyuQian (commented):
## Verdict

**Request changes.** Solid Phase 1 foundation — store/parser/resolver pipeline is well-decomposed and tests drive each task. Two correctness issues should be fixed before later phases depend on this graph's edges, plus repo-wide test convention drift that should be addressed now while the surface area is small.

## TL;DR

- **Should-fix:** C-heap leak (tree-sitter parser not closed); `implements` resolver matches struct methods across files when names collide.
- **Consider:** `tests` edges fan out to every callee in a test body (including `t.Fatal`, `fmt.Println`); import-path truncation to last segment risks future ID collisions; `isGoTestFunc` accepts `Testfoo`-style names that `go test` does not run.
- **Nit:** `Parser` interface doc-comment claim about concurrency safety is fragile.
- **CLAUDE.md compliance:** none of the new tests are table-driven with subtests, which the repo's CLAUDE.md, AGENTS.md, and GOLDEN_PRINCIPLES.md explicitly require.

## Strengths

- Clean two-phase architecture (parse → resolve) and stateless `Parser` interface aligns with the design spec §3.
- All exported (and most unexported) symbols carry proper Go doc comments starting with the symbol name.
- Empty-interface short-circuit in `addImplementsEdges` correctly avoids the every-concrete-implements-everything trap.
- SQL is parameterized everywhere; transactions wrap batch inserts; `INSERT OR REPLACE` / `INSERT OR IGNORE` chosen consciously per write semantics.
- `tools.go` build-tag pin for tree-sitter grammars is the standard Go community pattern.
- Fixtures are small, language-appropriate, and exercise both positive (Console implements Greeter) and negative (Mute does not) paths.

## Unknown-Unknowns Sweep

- **Q1 Local pattern fit:** ✅ Sub-packaging under `internal/graph/{parser,store,resolver}` mirrors `internal/auth`, `internal/generate`, etc. — no drift.
- **Q2 Blast radius:** ✅ Zero callers of any exported graph symbol outside `internal/graph/` itself. Confirms Phase 1's intentional isolation.
- **Q3 Pitfalls for this change class:** ⚠️ C-handle leak (sitter.Parser) confirmed; SQLite usage is safe (parameterized, WAL); malformed-source robustness is acceptable (tree-sitter returns partial trees, code does not dereference past `tree.RootNode()`).
- **Q4 Stale-training:** ✅ `modernc.org/sqlite v1.50.1` is current; `smacker/go-tree-sitter @ 2024-08-27` is the latest commit on main (no tagged release exists).
- **Q5 Hand-rolled vs off-the-shelf:** ✅ Name-matching resolver is justified vs `golang.org/x/tools/go/packages` — sub-millisecond and tolerant of unparseable repos, at the cost of accepting first-wins for duplicate names (documented).

## Counts

- Blocking: 0
- Should-fix: 2
- Consider: 3
- Nit: 1
- CLAUDE.md compliance: 1 (broad — all 4 new test files)

## Open Questions

1. Is the `tests`-edge spam (every callee in a test body) intentional for Phase 1 because filtering happens at query time, or should the parser narrow at emit time? Design spec §3 says `tests` represents "this test covers this function" which implies narrowing.
2. Should `external::` IDs keep the full import path (`external::github.com/foo/bar`) to be future-proof against monorepo collisions, accepting longer IDs? The schema would otherwise need a migration to fix this post-v1.
3. Are interface-method *signatures* (not just names) needed for v1, or is Phase 2's TS/Rust LSP cross-check sufficient to catch false positives? Spec §11 mentions "v1 restricts implements inference to exact method-set match" — confirming "method set" means names only here, not signatures.

internal/graph/store/store_test.go
```diff
@@ -0,0 +1,78 @@
...
+	t.Helper()
+	s, err := Open(t.TempDir() + "/graph.db")
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { s.Close() })
+	return s
+}
+
+func TestInsertAndGetNode(t *testing.T) {
```
SiyuQian: **CLAUDE.md compliance · Confidence 100**

`CLAUDE.md`, `AGENTS.md`, and `GOLDEN_PRINCIPLES.md` all state:

> Tests are table-driven with named subtests; don't mock our own packages.

None of the new tests in this PR are table-driven:
- `internal/graph/store/store_test.go` — `TestOpenCreatesSchema`, `TestInsertAndGetNode`, `TestCallersOf` are all single-case straight-line tests.
- `internal/graph/parser/parser_test.go` — single case.
- `internal/graph/parser/go_test.go` — 5 separate `TestGoParserExtracts*` functions, each single-case.
- `internal/graph/resolver/resolver_test.go` — 2 single-case tests.

This is the right time to address it — the test surface is still small and the team's convention should set the pattern for Phases 2-7 (TS/JS/Rust parsers, query layer) where each will have many more cases.

Fix pattern (per `GOLDEN_PRINCIPLES.md §9`):
```go
func TestGoParserExtracts(t *testing.T) {
    tests := []struct{
        name      string
        fixture   string
        wantNodes map[string]string // id -> kind
        wantEdges []store.Edge
    }{
        {name: "file_and_functions", fixture: "simple/main.go", ...},
        {name: "methods", fixture: "simple/main.go", ...},
        {name: "types", fixture: "simple/main.go", ...},
        {name: "calls", fixture: "simple/main.go", ...},
        {name: "imports", fixture: "simple/main.go", ...},
        {name: "tests_edges", fixture: "simple/main_test.go", ...},
    }
    for _, tc := range tests {
        t.Run(tc.name, func(t *testing.T) { ... })
    }
}
```

Anchored to this file as the most representative — applies to all four test files added by this PR.

internal/graph/parser/go.go
SiyuQian: Fixed in a872ef4 — `defer parser.Close()` immediately after `sitter.NewParser()`. Kept per-call allocation to preserve the concurrency contract.

internal/graph/parser/go.go
SiyuQian: Fixed in b124ed5 — `tests` edges now skip dsts of the form `external::pkg.Sym` (with a dot). Unresolved intra-module names (`external::Name` no dot) are kept so the resolver can rewrite them. Existing test still passes because the fixture `Greet` is bare-name.

internal/graph/parser/go.go
SiyuQian: Fixed in b246f8c — added the `unicode.IsLower([]rune(name)[4])` check after the prefix test. Doc comment updated to spell out the rule.

internal/graph/parser/go.go
SiyuQian: Fixed in a45f472 — `importSpecPath` now returns the full path. `lastSlash` helper removed. Existing test still passes because the fixture imports ("fmt", "strings") have no slashes.

internal/graph/resolver/implements.go
SiyuQian: Fixed in ae8e332 — changed to an exact lookup `structs[n.Path+"::"+n.Container]`. Documented in code that multi-file methods within a Go package are a known v1 false-negative (conservative trade), to be revisited when the package resolver lands.

internal/graph/parser/parser.go
SiyuQian: Fixed in ca4d222 — expanded the comment to make the contract explicit (no shared mutable state across Parse calls; tree-sitter wrappers must allocate fresh per call OR add explicit sync).

internal/graph/store/store_test.go
SiyuQian: Fixed in 70c048b — all four new test files converted to table-driven with named subtests. Subtest count went from 12 flat tests to 20 named subtests (`TestStoreSchema` expanded to 3 per-table cases; `TestGoParserExtracts` collapsed 6 funcs into 6 subtests; `TestResolve` has 2 subtests). `go test ./internal/graph/... -count=1 -v` shows 20 PASS.

PR #109, opened 2026-05-20
feat(graph): Phase 2 — TS/JS/Rust parsers + cache layer + incremental builder

SiyuQian:
## Summary

Phase 2 of the devpilot graph subsystem (PR #108 was Phase 1). Implements all 12 sub-tasks from `docs/plans/2026-05-19-devpilot-graph-plan.md` §Phase 2 across 28 commits tagged `graph-phase-2`. Bite-sized expansion lives in `docs/plans/2026-05-20-devpilot-graph-phase2-plan.md`.

**Adds (≈6900 LoC, 41 files):**
- **TypeScript / JavaScript / Rust parsers** under `internal/graph/parser` using `github.com/smacker/go-tree-sitter` bindings. Each emits the same `parser.ParseResult` shape as Phase 1's Go parser (file/function/method/class/interface/type nodes + `contains`/`calls`/`imports`/`tests`/`implements`/`extends` edges).
- **tsconfig path-alias resolver** (`internal/graph/resolver/tsconfig.go`) — handles `extends` chain + `baseUrl` + `paths` and rewrites `external::<alias>` import edges to repo-relative paths.
- **Cache package** (`internal/graph/cache`):
  - `RepoKey(absRepoRoot)` → deterministic 12-char sha1 hex
  - On-disk layout helpers under `~/.devpilot/graphs/<key>/`
  - `gofrs/flock` build lock with 60s timeout
  - TTL sweeper for `~/.devpilot/preflight/` (7-day default)
  - Atomic `meta.json` read/write (write-temp-then-rename)
  - `parser.DefaultRegistry()` mapping extensions to parsers
  - `Builder` with sequential→parallel parse (`errgroup`, deterministic indexed results)
  - `Build()` dispatcher: missing graph.db → `FullBuild`; schema mismatch → wipe + `FullBuild`; else `BuildIncremental`
  - `BuildIncremental` via `git diff --name-status` with full-rebuild fallback when prev SHA is not ancestor of HEAD
  - Lazy preflight TTL sweep on every `Build()`
- **Phase 2 acceptance harness** (`internal/graph/cache/phase2_acceptance_test.go`) with 4 subtests — all green.

## Deviations from plan

- **Resolver API**: plan-text pseudocode referenced `resolver.New(nodes).Rewrite(edges)` and `resolver.Implements(...)` which never existed. Implementation uses the actual `resolver.Resolve([]parser.ParseResult) []parser.ParseResult` entrypoint (which internally handles both external-edge rewrite and implements-edge derivation). For incremental builds, existing nodes from `store.AllNodes()` are wrapped into a synthetic `ParseResult` and prepended so the resolver can see them.
- `unquote` helper landed in a new `internal/graph/parser/strings.go` rather than the plan's suggested `tools.go` (which has a `//go:build tools` tag excluding it from normal builds).

## Known follow-ups (non-blocking, out of plan scope)

Flagged by code reviewers during execution; recorded here for Phase 3 hardening:
- TS: arrow-const declarations (`export const handler = () => {}`) not added to intra map
- TS: `new Foo()` (`new_expression`) not treated as a call
- TS: intra-class method calls (`this.foo()`) emit `external::this.foo` rather than resolving to sibling methods
- tsconfig: JSONC `//` comments inside string values get stripped naively; no `/* */` block-comment support; no `extends` cycle protection; package-name `extends` (`@tsconfig/node20`) not resolved
- Rust: `flattenUseTree` doesn't expand `use foo::{a, b};` grouped imports or `use_as_clause` aliases

## Review Guide

Suggested order for reviewers:
1. `docs/plans/2026-05-20-devpilot-graph-phase2-plan.md` — the bite-sized plan that drove execution
2. `internal/graph/parser/typescript.go` + `typescript_test.go` — sets the pattern; JS and Rust parsers follow it
3. `internal/graph/resolver/tsconfig.go` — the only non-mechanical resolver
4. `internal/graph/cache/builder.go` + `incremental.go` — the integration point that ties parser+resolver+store together; this is where the resolver-API drift forced the biggest deviation
5. `internal/graph/cache/phase2_acceptance_test.go` — the four acceptance criteria

Watch for:
- Determinism: `Builder.FullBuild` must produce byte-identical `graph.db` on consecutive runs and across parallel vs sequential. `parser.Registry.Languages()` sorts its output for this reason.
- `BuildIncremental`'s edge handling: it deletes nodes/edges for changed-or-deleted files, re-parses added+modified, then runs the resolver against the *union* of existing-store nodes + new ParseResults so cross-file resolution still works.

## Test plan

- [x] `make test` — all packages green
- [x] `make lint` — 0 issues
- [x] `TestPhase2Acceptance` — 4/4 subtests pass (RepoKey, two-builds-identical, incremental-matches-full, flock-serializes)
- [ ] Manual: try building the devpilot repo itself (`devpilot graph build .` — wait, that CLI doesn't exist yet, Phase 4). Skip until CLI lands.

## Out of scope (subsequent phases)

- Phase 3: query layer (`CallersOf`, `Preflight`, etc.)
- Phase 4: `devpilot graph <verb>` CLI surface + JSON envelope
- Phase 5: LSP cross-check test infrastructure
- Phase 6: skill integration (`preflight.sh`, devpilot-pr-review wiring)
- Phase 7: polish + release gate


SiyuQian (commented):
## Verdict: Comment

Phase 2 delivers a coherent parser + cache subsystem and the Phase 2 acceptance harness is green. Most findings here are **fixable correctness/operability issues** found by the fanout (5 should-fix, 5 consider, 4 nit, 1 style). None block merge on their own; the bundle is worth a follow-up pass before Phase 3 layers query/CLI on top.

## TL;DR

- Three correctness issues threaten the documented determinism guarantee: tsconfig `paths` map iterated in random order, JSONC stripper corrupts `//` inside string values, and FullBuild wipes `graph.db` without its WAL/SHM sidecars.
- One convention drift: Rust parser fabricates synthetic `path::name` IDs for unresolved identifier calls, while Phase 1/Go and TS both use `external::name` and rely on the resolver to rewrite. The Rust comment claims resolver compatibility but the resolver only rewrites the `external::` form.
- Incremental update has two operability gaps worth knowing about: it only sees committed changes (silent staleness from dirty trees) and uses `CombinedOutput` + `strings.Fields` on git output (stderr noise + paths-with-spaces fragility).
- CLAUDE.md rules: cache/resolver tests skip the table-driven convention the parser tests follow; several cache-layer error returns propagate `err` bare instead of wrapping at the boundary.

## Strengths

- TS / JS / Rust parsers consistently allocate a fresh `sitter.Parser` per `Parse` call with `defer Close()` — Phase 1's C-heap leak (caught in PR #108 review) is not reintroduced. Verified across all three new parsers.
- Determinism invariant on the parallel parse path is well-stated (`builder.go:91` comment is exactly the kind of "DO NOT swap to channel/map" note future agents will want). `parser.Registry.Languages()` sorts its output for this reason. `TestBuilderFullBuildDeterministic` and the parallel-equals-sequential test pin it down.
- Plan-to-API drift handled gracefully: the bite-sized plan referenced a `resolver.New().Rewrite()` API that never existed; implementation correctly adapted to Phase 1's `resolver.Resolve([]ParseResult)` and called the deviation out in the PR body.
- Phase 2 acceptance harness covers the four spec criteria and exercises the `git diff` incremental path end-to-end.
- `tsconfig.go` is a clean, small resolver with `extends` chain support — easy to reason about.

## Unknown-Unknowns Sweep (Agent A)

1. **Local pattern fit**: parsers mirror Phase 1 `go.go` shape (NewParser/Close, ParseCtx, ParseResult); cache layer is new with consistent error-wrapping pattern (where it wraps — see Agent C).
2. **Blast radius**: only `cache.Builder` and `cache.BuildIncremental` consume `parser.Registry` + `store.Store` today; no Phase 3/4 CLI yet. New `Store.AllNodes` and `Store.DeleteByPaths` are sound additions.
3. **Pitfall coverage**: build lock present and tested; SQL parameterized; SQLite WAL pragma set BUT WAL/SHM sidecars not cleaned up on FullBuild wipe; `git diff` parsing fragile against stderr + spaces; no tree-sitter parse timeout.
4. **Stale-training**: go 1.25.6 + tree-sitter pin + modernc/sqlite v1.50.1 + gofrs/flock v0.13.0 all current; nothing odd.
5. **Hand-rolled vs OTS**: `stripJSONComments` is a naive line-scan; the Go ecosystem has `github.com/tailscale/hujson` and similar that handle the JSONC-in-strings case correctly.

## Counts

| Severity | Count |
|---|---|
| Should-fix | 5 |
| Consider | 5 |
| Nit | 4 |
| Style / CLAUDE.md | 2 |

**Total inline comments:** 16. All anchored to lines this PR modified.

## Open Questions

1. The `external::` ID convention seems load-bearing for resolver rewrites. Is `rust.go:120`'s `filePath + "::" + name` form intentional (and resolver-handled somewhere I missed), or did the convention drift accidentally?
2. Should `BuildIncremental` walk dirty working-tree files (`git status --porcelain`), or should it require a clean tree and bail to FullBuild otherwise? Both are reasonable; pick one and document.
3. Phase 3 query layer will read every edge as ground truth — should we filter `external::this.foo`-shaped pollution at insert time, or leave it for `query/preflight.go` to ignore?

## What I did NOT review

- The known follow-ups listed in the PR description (arrow-const TS calls, `new Foo()`, intra-class method resolution, JSONC-in-strings, extends cycle protection, Rust grouped-use) are deliberately deferred — not re-flagged here as blockers. Some of these (cycle protection, JSONC) ended up surfacing as `Consider` findings below because I think they're worth doing before the cache ships to real repos, not Phase 3.

internal/graph/resolver/tsconfig.go
```diff
@@ -0,0 +1,118 @@
...
+					}
+					return filepath.ToSlash(rel), true
+				}
+			}
+		}
+	}
+	return "", false
+}
+
+func stripJSONComments(b []byte) []byte {
```
SiyuQian, line 138: **Should-fix · Confidence 95** — `stripJSONComments` truncates at the first `//` on each line, including inside string values. Any tsconfig containing a URL (`"jsxImportSource": "https://..."`) or a path with a literal `//` gets silently corrupted before `json.Unmarshal`, producing either a parse error or silently-missing fields — and then alias resolution is disabled for that repo, polluting the cache with unresolved `external::` edges.

Fix: track string state while stripping, or swap for `github.com/tailscale/hujson` (standardize, then unmarshal). Repro is one line: `{"compilerOptions": {"baseUrl": "https://x"}}` → empty `baseUrl` after stripping.

internal/graph/resolver/tsconfig.go
```diff
@@ -0,0 +1,118 @@
...
+		spec := strings.TrimPrefix(e.Dst, "external::")
+		if rel, ok := r.resolve(spec); ok {
+			out[i].Dst = rel
+		}
+	}
+	return out
+}
+
+func (r *TSConfigResolver) resolve(spec string) (string, bool) {
+	for pattern, targets := range r.paths {
```
SiyuQian: **Should-fix · Confidence 85** — `r.paths` is a Go map; iteration order is randomized per process. When two patterns can match the same spec (e.g. `@app/*` and `@app/foo/*`), the resolved file path varies run-to-run.

This silently breaks the determinism guarantee that `Builder.FullBuild` advertises and that `TestBuilderFullBuildDeterministic` asserts via dump equality — the existing test won't catch it because the fixture only has one alias.

Fix: snapshot `keys := slices.Sorted(maps.Keys(r.paths))` (longest-prefix first, then lexical) and iterate that.

internal/graph/resolver/tsconfig.go
```diff
@@ -0,0 +1,118 @@
...
+// NewTSConfigResolver loads tsconfig.json from root (handling `extends` once).
+func NewTSConfigResolver(root string) (*TSConfigResolver, error) {
+	r := &TSConfigResolver{root: root, paths: map[string][]string{}}
+	if err := r.loadTSConfig(filepath.Join(root, "tsconfig.json")); err != nil {
+		return nil, err
+	}
+	return r, nil
+}
+
+func (r *TSConfigResolver) loadTSConfig(path string) error {
```
SiyuQian: **Consider · Confidence 85** — `loadTSConfig` recurses on `cfg.Extends` with no visited set. A circular extends chain (`A.json extends B.json extends A.json`) recurses until stack overflow and crashes the whole build, not just the resolver.

Not a hypothetical in shared-config monorepos. Fix: pass a `visited map[string]bool` through and return an error (or `nil`) on revisit. The PR description acknowledges this as a known follow-up but it's cheap enough to do here.

internal/graph/resolver/tsconfig.go
```diff
@@ -0,0 +1,118 @@
+// Package resolver — additions for TypeScript path aliases.
```
SiyuQian: **Nit · Confidence 90** — `// Package resolver — additions for TypeScript path aliases.` adds a second package doc comment to the `resolver` package; `resolver.go` already carries the canonical one ("Package resolver rewrites synthetic external:: edges...").

Go style expects exactly one package comment per package. `go doc ./internal/graph/resolver` will pick one nondeterministically. Drop this line or fold the TSConfigResolver description into `resolver.go`'s package doc.

internal/graph/cache/builder.go
```diff
@@ -0,0 +1,186 @@
...
+// first so two consecutive calls produce identical output.
+func (b *Builder) FullBuild() (BuildResult, error) {
+	rel, err := AcquireBuildLock(LockFile(b.home, b.key), 60*time.Second)
+	if err != nil {
+		return BuildResult{}, err
+	}
+	defer func() { _ = rel() }()
+
+	dbPath := GraphDB(b.home, b.key)
+	if err := os.Remove(dbPath); err != nil && !os.IsNotExist(err) {
```
SiyuQian: **Should-fix · Confidence 90** — `os.Remove(dbPath)` wipes `graph.db` but leaves the SQLite WAL sidecars (`graph.db-wal`, `graph.db-shm`) that `store.Open` creates via `journal_mode=WAL`. After a crashed build, the next `FullBuild` opens a fresh empty `graph.db` next to a stale WAL — SQLite may replay it on the empty db, reintroducing deleted rows or causing `database is locked` errors.

Fix: also remove `dbPath+"-wal"` and `dbPath+"-shm"`, tolerating `os.IsNotExist`. Two extra `os.Remove` calls.

internal/graph/cache/builder.go
```diff
@@ -0,0 +1,186 @@
...
+		Mode:        "full",
+	}, nil
+}
+
+// Build dispatches to FullBuild or BuildIncremental based on cache state.
+// Missing graph.db -> full. Schema-version mismatch -> wipe cache dir, full.
+// Otherwise -> incremental (currently stubbed to full; see Task 2.21).
+func (b *Builder) Build() (BuildResult, error) {
+	_ = SweepPreflight(b.home, 7*24*time.Hour)
+	if _, err := os.Stat(GraphDB(b.home, b.key)); os.IsNotExist(err) {
```
SiyuQian, line 169: **Consider · Confidence 80** — TOCTOU between `os.Stat(GraphDB)` and `AcquireBuildLock` inside `FullBuild`. Two concurrent `Build()` calls can both see "missing db", both proceed to `FullBuild`, the second one then rebuilds from scratch rather than reading the cache just produced.

Correctness is preserved (the lock inside `FullBuild` serializes them), but the lock's point is to avoid that redundant work. Fix: acquire the lock at the top of `Build()`, then re-stat under the lock and dispatch.

internal/graph/cache/builder.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("abs(%s): %w", repo, err)
+	}
+	info, err := os.Stat(abs)
+	if err != nil || !info.IsDir() {
+		return nil, fmt.Errorf("repo %s is not a directory", abs)
+	}
+	key := RepoKey(abs)
+	if err := EnsureDirs(home, key); err != nil {
+		return nil, err
```
SiyuQian, line 43: **Consider · Confidence 80 · CLAUDE.md** — `CLAUDE.md`: *"Wrap errors at layer boundaries: `fmt.Errorf("doing X: %w", err)`."*

This call site (and several others in `builder.go` / `incremental.go`) returns store/resolver/parser errors bare:
```go
return BuildResult{}, err
```
The cache package IS the layer boundary against parser/resolver/store — these should wrap with the operation name (`"open graph.db"`, `"insert nodes"`, `"parse fanout"`, `"acquire build lock"`, etc.) so failed-build logs say *what stage* failed, not just the underlying error. The same file already does this correctly at builder.go:68/73/101; the missing wraps are local inconsistency.

internal/graph/cache/incremental.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/siyuqian/devpilot/internal/graph/store"
+)
+
+// BuildIncremental applies a delta on top of an existing graph.db by diffing
+// prev.HeadSHA against the current HEAD with `git diff --name-status`. It
+// falls back to FullBuild when prev.HeadSHA is empty, the git invocation
+// fails, or prev.HeadSHA is not an ancestor of HEAD (force-push, rebase).
+func (b *Builder) BuildIncremental(prev Meta) (BuildResult, error) {
+	currentHead := gitHeadSHA(b.repo)
+	if prev.HeadSHA == "" || currentHead == "" || !isAncestor(b.repo, prev.HeadSHA, currentHead) {
```
SiyuQian, line 24: **Consider · Confidence 85** — `BuildIncremental` derives changes from `git diff prev.HeadSHA HEAD` only, so files modified in the working tree but not yet committed are silently ignored. When invoked from a dirty checkout (typical local dev or pre-commit hooks), the cache reflects the last-committed content while `FullBuild` would walk the working tree — users see preflight outputs that disagree with the file on disk.

Pick a posture:
- Union `git status --porcelain` files into `Modified`/`Added`, OR
- Detect dirty tree and bail to `FullBuild`, OR
- Document that incremental is committed-state-only.

Third option is fine but should be in the package doc.

internal/graph/cache/incremental.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+type changeSet struct {
+	Added, Modified, Deleted []string
+}
+
+// gitChangedFiles parses `git diff --name-status from to` into added,
+// modified, and deleted path slices. Renames are split into delete+add.
+func gitChangedFiles(repo, from, to string) (changeSet, error) {
+	out, err := exec.Command("git", "-C", repo, "diff", "--name-status", from, to).CombinedOutput()
```
SiyuQian: **Should-fix · Confidence 85** — Two fragility bugs in one line:

1. `CombinedOutput` mixes stderr warnings (`warning: in the working copy of X, LF will be replaced by CRLF`) into the diff. The parser below then sees those lines as status records.
2. `strings.Fields(line)` splits on any whitespace; paths containing spaces (`src/My App.ts`) get fragmented and the A/M/D branches capture only the trailing token as the path.

Fix: use `exec.Output()` (separate stderr), and pass `-z` for NUL-separated output:
```go
exec.Command("git", "-C", repo, "diff", "--name-status", "-z", from, to).Output()
```
Then split on `\x00` and parse name-status records explicitly. The rename branch (`fields[1]`/`fields[2]`) also benefits because `-z` separates source and dest into their own records.

internal/graph/cache/incremental.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+		status := fields[0]
+		switch status[0] {
+		case 'A':
+			cs.Added = append(cs.Added, filepath.ToSlash(fields[len(fields)-1]))
+		case 'M':
+			cs.Modified = append(cs.Modified, filepath.ToSlash(fields[len(fields)-1]))
+		case 'D':
+			cs.Deleted = append(cs.Deleted, filepath.ToSlash(fields[len(fields)-1]))
+		case 'R':
```
SiyuQian: **Consider · Confidence 80** — switch on `status[0]` covers `A`/`M`/`D`/`R` only. `C` (copy), `T` (typechange — file became a symlink/etc.), and `U` (unmerged) silently drop the file. The `R` branch additionally hard-codes `len(fields) == 3`, which fails with `-z` (no fields) or when paths contain spaces (>3 fields).

Safe default: treat unknown statuses as `Modified` so the file gets re-parsed rather than skipped. Otherwise the cache drifts from reality on copy/typechange edits.

internal/graph/cache/meta.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	data, err := json.MarshalIndent(m, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshal meta: %w", err)
+	}
+	tmp := path + ".tmp"
+	if err := os.WriteFile(tmp, data, 0o644); err != nil {
+		return fmt.Errorf("write %s: %w", tmp, err)
+	}
+	if err := os.Rename(tmp, path); err != nil {
```
SiyuQian, line 53: **Consider · Confidence 75** — `os.WriteFile` + `os.Rename` is atomic against other readers but not against power loss: depending on the FS the rename may resolve before the file's data is on stable storage, leaving an empty/truncated `meta.json`.

The blast radius is mild — `ReadMeta` returns zero `Meta`, `Build()` sees `SchemaVersion=0 ≠ CurrentSchemaVersion`, and we full-rebuild. So this is more an operability note than a correctness bug. If you want hard durability: open tmp, Write, Sync, Close, Rename, then `fsync(filepath.Dir(path))`. Otherwise add a comment that crash-during-meta-write triggers a full rebuild on recovery.

internal/graph/cache/flock.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/gofrs/flock"
+)
+
+// ErrLockTimeout indicates AcquireBuildLock could not obtain the lock within timeout.
+var ErrLockTimeout = errors.New("build lock acquire timed out")
+
+// AcquireBuildLock takes an exclusive flock on lockPath, polling every 100ms until
+// timeout. Returns a release function the caller must invoke.
+func AcquireBuildLock(lockPath string, timeout time.Duration) (release func() error, err error) {
```
SiyuQian, line 19: **Nit · Confidence 90** — `flock.New(lockPath)` will hit `ENOENT` on the first call if the parent directory doesn't exist. Today `NewBuilder` calls `EnsureDirs` first so this is fine, but the helper is callable independently (Phase 4 CLI is the obvious upcoming user).

One-liner: `_ = os.MkdirAll(filepath.Dir(lockPath), 0o755)` at the top of `AcquireBuildLock`. Defensive, cheap.

internal/graph/cache/walk.go
```diff
@@ -0,0 +1,54 @@
...
+	"io/fs"
+	"path/filepath"
+	"strings"
+)
+
+// WalkRepo returns every regular file in root, skipping VCS and dependency
+// directories. Paths returned are repo-relative with forward slashes.
+func WalkRepo(root string) ([]string, error) {
+	var out []string
+	skipDirs := map[string]bool{
```
SiyuQian, line 13: **Nit · Confidence 85** — `skipDirs` covers `.git`, `node_modules`, `target`, `vendor`, `.devpilot` but misses common build-output dirs: `dist`, `build`, `out`, `.next`, `.turbo`, `.cache`, `coverage`, `__pycache__`.

For any repo with TS source → JS build output sitting next to it, every output file gets re-parsed, doubling node counts and creating bogus "duplicate definition" name-collisions that the resolver's first-occurrence-wins index will silently mis-resolve.

Fix: extend the skip set, or (better) load `.gitignore` and use a matcher. The former is one line and covers ~95% of cases.

internal/graph/parser/rust.go
```diff
@@ -0,0 +1,235 @@
...
+		if n.Type() == "call_expression" {
+			fn := n.ChildByFieldName("function")
+			if fn != nil {
+				switch fn.Type() {
+				case "identifier":
+					name := fn.Content(src)
+					if id, ok := intra[name]; ok {
+						out = append(out, store.Edge{Src: srcID, Dst: id, Kind: "calls"})
+					} else {
+						out = append(out, store.Edge{Src: srcID, Dst: filePath + "::" + name, Kind: "calls"})
```
SiyuQian: **Should-fix · Confidence 85** — Convention drift from Phase 1 / TS. The walker emits `filePath + "::" + name` for unresolved identifier calls:
```go
out = append(out, store.Edge{Src: srcID, Dst: filePath + "::" + name, Kind: "calls"})
```
But `internal/graph/parser/go.go:80-82` documents the convention as: *"Unresolved intra-module names (external::Name with no dot) are kept so the resolver can rewrite them to real node IDs."* The resolver (`internal/graph/resolver/resolver.go`) only rewrites the `external::Name` form — it does NOT rewrite arbitrary `path::name` edges.

Consequence: every Rust identifier call to a function defined in another file produces a `<thisfile>::name` edge that points at a node that doesn't exist (the function lives at `<otherfile>::name`). The resolver leaves it alone, and the query layer (Phase 3) joins against a dangling Dst.

Fix: emit `external::name` like the Go/TS parsers do (`internal/graph/parser/typescript.go:271` for the analogue). Phase 1's resolver handles the rewrite for free.

internal/graph/parser/typescript.go
```diff
@@ -0,0 +1,335 @@
...
+	var out []store.Edge
+	var visit func(n *sitter.Node)
+	visit = func(n *sitter.Node) {
+		if n == nil {
+			return
+		}
+		if n.Type() == "call_expression" {
+			fn := n.ChildByFieldName("function")
+			if fn != nil && fn.Type() == "identifier" {
+				out = append(out, store.Edge{Src: path, Dst: "external::" + fn.Content(src), Kind: "tests"})
```
SiyuQian, line 144: **Nit · Confidence 70** — `extractCallTargets` always emits `external::<ident>` for `tests` edges, bypassing the intra-file map that the sibling `walkTSCalls` (~line 245) consults. So a Jest test like `it("...", () => greet("x"))` where `greet` is in the same file produces `tests` → `external::greet` instead of `tests` → `<path>::greet`.

Go's analogous test-edge handling in `internal/graph/parser/go.go:77-85` deliberately preserves the intra-resolved form so Phase 1's resolver can rewrite. Either pass `intra` through to `extractCallTargets`, or have the resolver also rewrite `external::Name` for `tests`-kind edges (which it does, but only if the call site emits the `external::` form to begin with). The current code emits `external::greet` which the resolver will pick up — so the wire effect is OK on second look, but it bypasses the intra-fast-path and diverges from how the Go parser handles the same case.

internal/graph/parser/typescript.go
```diff
@@ -0,0 +1,335 @@
...
+// NewTypeScriptParser returns a Parser for TypeScript source files.
+func NewTypeScriptParser() *TypeScriptParser {
+	return &TypeScriptParser{lang: tsLang.GetLanguage()}
+}
+
+func (p *TypeScriptParser) Language() string     { return "typescript" }
+func (p *TypeScriptParser) Extensions() []string { return []string{".ts", ".tsx"} }
+
+func (p *TypeScriptParser) Parse(path string, src []byte) (ParseResult, error) {
+	parser := sitter.NewParser()
```
SiyuQian, line 26: **Consider · Confidence 70** — `parser.ParseCtx(context.Background(), nil, src)` never cancels. A pathological/minified input (e.g. a 30 MB bundled `.js` that slipped through the skip list, or a tree-sitter grammar edge case) can pin a parser goroutine indefinitely. With `errgroup.SetLimit(4)` four such files stall the whole build until the 60s flock timeout fires somewhere else.

Fix: plumb a per-file timeout (30s? 10s?) via `context.WithTimeout`. On `ctx.Err`, treat as a soft skip with a `ParseError` rather than failing the whole build.

PR #110, opened 2026-05-20
feat(graph): Phase 3 — query layer + preflight composite

SiyuQian:
## Summary

Phase 3 of the devpilot-graph subsystem: the read-side query layer that the upcoming CLI (Phase 4) and `devpilot-pr-review` skill (Phase 6) will consume. No CLI surface yet — this PR only lands the library.

New package `internal/graph/query` with:

- **`Reader`** — narrow read-only interface (`reader.go`), satisfied by `*store.Store`.
- **Primitive queries** — `CallersOf`, `CalleesOf` (depth-limited BFS); `TestsFor`, `ImplementorsOf` (direct edge lookup); `Hubs` (group-by inbound calls); `ImpactRadius` (file → symbol union of callers); `Context` (source snippet + caller snippets, reads from disk).
- **`DetectChanges`** — shells out to `git diff --name-status -M` + `git show`; for modified files joins with the graph by path, emits one entry per symbol when blob sha256 differs; for A/D/R emits a file-level entry. The `git` call is routed through a package-level `gitRun` variable so tests stub it.
- **`RiskScore`** — pure formula from design §6 (`2*exported + 3*hub + 3*iface + 1*untested`).
- **`Preflight`** composite — wires `DetectChanges` + enrichment + hub lookup + cross-community aggregation + ranking + truncation; returns the §6 JSON shape verbatim (verified by `preflight_roundtrip_test.go`).

Store extensions to support the above: `EdgesByDst`, `EdgesBySrc`, `NodesByPath`, `CountEdgesByKind`, `HubsByCalls`.

Plan: \`docs/plans/2026-05-20-devpilot-graph-phase3-plan.md\` (also added in this PR — the bite-sized expansion of master plan Phase 3).

## Review Guide

Suggested reading order:

1. **\`docs/plans/2026-05-20-devpilot-graph-phase3-plan.md\`** — the contract this PR implements; every commit corresponds to one numbered task.
2. **\`internal/graph/store/store.go\`** — the five new accessors. They're trivial SQL but determine the shape of the `Reader` interface.
3. **\`internal/graph/query/reader.go\`** — the read-only surface every query depends on.
4. **\`internal/graph/query/callers.go\` + \`callees.go\`** — the BFS pattern; everything else either reuses these or is a one-shot lookup.
5. **\`internal/graph/query/preflight.go\`** — the only non-trivial file (~380 lines). Read top-to-bottom: types → `communityFromPath` → `enrichChangedSymbol` → `Preflight` composite → `crossCommunityEdges` / `buildRiskSummary`.
6. **\`internal/graph/query/detect_changes.go\`** — the one place we shell out to git. The `gitRun` var is the seam for tests.

Things to watch:
- \`Preflight\` field tags map 1:1 to the design doc §6 JSON schema; \`Risk\` is the lone exception (\`json:"-"\`, used only for ranking).
- \`communityFromPath\` is capped at depth 3 by spec; the test pins this behavior.
- Cycle safety: \`CallersOf\` / \`CalleesOf\` use a `seen` set seeded with the target ID, so the target is never re-emitted and cycles terminate.
- BFS is hop-ordered, not insertion-ordered — tests sort before comparing.
- \`Context\` reads files from disk relative to \`repoRoot\`; missing files surface as wrapped errors except for caller snippets (silently skipped — synthetic \`external::*\` callers have no node).

## Test plan

- [x] \`go test ./internal/graph/...\` — 109 tests across 5 packages, all green
- [x] \`make lint\` — 0 issues
- [x] Each query verb in design §6 has at least one dedicated test file
- [x] \`preflight_roundtrip_test.go\` asserts the JSON marshalled from \`PreflightResult\` contains every field in the design §6 schema
- [x] Phase 4 CLI commands will consume these queries via thin wrappers — not in this PR
- [x] LSP cross-check (Phase 5) and skill integration (Phase 6) come later

### DIFF
--- a/docs/plans/2026-05-20-devpilot-graph-phase3-plan.md
+++ b/docs/plans/2026-05-20-devpilot-graph-phase3-plan.md
@@ -1,0 +1,2301 @@
+# Devpilot Graph Phase 3 — Query Layer Bite-Sized Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Implement every read-side query that downstream consumers (CLI in Phase 4, `devpilot-pr-review` skill in Phase 6) will use, including the `preflight` composite. All queries operate against a populated `*store.Store`; no parsing or git mutation happens here (except `DetectChanges`/`Preflight` which shell out to `git` read-only).
+
+**Architecture:** A new `internal/graph/query` package. Each verb is a single file with one exported function plus tiny internal helpers. Queries take a `*store.Store` (or a narrower interface defined locally) and primitive args; they return typed Go structs that map 1:1 to the Phase 4 JSON envelope shape. Composite queries (`Preflight`) call other query functions in this same package — no cycles, since `Preflight` lives in its own file. `DetectChanges` shells out to `git diff --name-status` and `git show :path` to read base/head trees, then joins with the graph by file path.
+
+**Tech Stack:**
+- Go 1.25.6, module `github.com/siyuqian/devpilot`
+- `modernc.org/sqlite` (via existing `internal/graph/store`)
+- `os/exec` for `git diff --name-status` + `git show`
+- Stdlib `crypto/sha256` for signature hashes
+
+**Conventions inherited from Phases 1–2:**
+- One passing test → one commit (`feat(graph/query): <verb> <noun>` for new behavior; `test(graph/query): ...` for tests-only).
+- All `fmt.Errorf` at layer boundaries wraps with `%w`.
+- Tests are table-driven with named subtests where >1 case; small fixtures are built via `store.InsertNodes` / `store.InsertEdges` directly (do NOT spin up parsers in query tests — keep them hermetic).
+- No mocking of our own packages. The single exception is `detect_changes.go`: factor the `git` invocation through a function variable so tests can swap it.
+- Queries MUST NOT mutate the store; any test that needs to assert no-write should use a `t.TempDir()`-backed store and re-open read-only after.
+- Run `make lint` and `make test` before committing each task.
+
+---
+
+## File Structure
+
+```
+internal/graph/query/
+├── reader.go               (new — narrow read-only interface satisfied by *store.Store)
+├── reader_test.go          (new — interface compile check)
+├── callers.go              (new — CallersOf with depth)
+├── callers_test.go         (new)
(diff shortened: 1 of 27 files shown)


SiyuQian:
### Code review

Found 2 issues:

1. `DetectChanges` swallows `git show` errors and emits misleading change types. When `git show <base>:<path>` fails (e.g. file added at head, no rename detected upstream), `baseBlob` is silently set to `nil`. Because `hashBytes(nil) == ""` differs from the head hash, every graph-known symbol in that file is then emitted as `change_type: "modified"` — directly contradicting the inline comment on line 78 ("treat as added"). The symmetric case (`git show <head>:<path>` fails) marks symbols as modified instead of surfacing a real error. Tests in `detect_changes_test.go` never exercise an error from `gitRun`, so the bug is invisible to CI.

https://github.com/SiyuQian/devpilot/blob/53110562a278526aaf12a4fce32680e39f80cf8d/internal/graph/query/detect_changes.go#L74-L90

2. `crossCommunityEdges` aggregates over the sampled callers, not all callers. `enrichChangedSymbol` truncates `Callers.Sample` to `CallerSample` (default 10), but `crossCommunityEdges` iterates only that sample (line 296), and `RiskSummary.NewCrossCommunityEdges` is the sum of those `CountAdded` increments (line 344). For any modified symbol with >10 callers, both the per-edge `count_added` and the summary `new_cross_community_edges` will silently under-count. The field name `CountAdded` also implies "edges added in this diff" but is really "sampled callers crossing communities".

https://github.com/SiyuQian/devpilot/blob/53110562a278526aaf12a4fce32680e39f80cf8d/internal/graph/query/preflight.go#L287-L346

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

SiyuQian:
Fixed in 39d1667e1ed4f68d974aabdac6bda38822db5d76.

1. `DetectChanges` now classifies the rename edge case correctly: when one side's blob is missing it emits `added` / `removed` as a file-level entry instead of falsely marking every graph symbol as `modified`. If both sides fail, the git error is surfaced. Regression test: `TestDetectChangesRenameEdgeCase`.
2. `crossCommunityEdges` now iterates all inbound `calls` edges via `EdgesByDst` rather than the truncated `Callers.Sample`, so `count_added` and `risk_summary.new_cross_community_edges` are no longer capped by `CallerSample`. Regression test `TestCrossCommunityEdgesCountsAllCallers` uses 12 callers with a sample of length 1 and asserts `count_added == 12`.

All 111 graph tests pass, lint clean.

PR #112, opened 2026-05-21
feat(graph): Phase N1 — native Go AST backend behind env flag

SiyuQian:
## Summary

Phase N1 of the [native-AST migration](docs/plans/2026-05-20-graph-native-ast-design.md): adds a `go/packages` + `go/types` Go parser backend behind the `DEVPILOT_GRAPH_GO_BACKEND=native` env flag. Default remains tree-sitter for one release while N1 bakes. 16 half-day tasks across 20 commits; all-green CI + lint.

Closes the gap from the original Phase 1–4 work where the Go resolver used **name-only heuristics** (first-occurrence wins for any `external::Foo` placeholder) and `implements` fell back to method-name-set matching. With native, every `calls` / `implements` / `tests` / `imports` edge is resolved via `types.Info.Uses` / `types.Implements` against the loaded module.

## What changed

- **`internal/graph/parser/go_native.go`** (~700 lines, new) — `GoNativeParser` implements both `Parser` (no-op `Parse`) and the new additive `PackageLoader` interface. `LoadModule(repoRoot)` runs `packages.Load` once with `NeedName|NeedFiles|NeedImports|NeedDeps|NeedTypes|NeedSyntax|NeedTypesInfo`, then walks every file via three passes (mint nodes + index → call/test edges → implements).
- **`internal/graph/parser/parser.go`** — new `PackageLoader` interface (additive; existing parsers unchanged).
- **`internal/graph/parser/registry.go`** — env-flag-driven backend selection via `goBackendFromEnv()`; new `Registry.ForLanguage(lang)` and `Registry.GoBackend()` helpers.
- **`internal/graph/cache/native_go.go`** (new) — `loadGoModule` with `go.work` (`modfile.ParseWork`) + `go.mod` + non-module sentinel (`errNoGoModule`).
- **`internal/graph/cache/builder.go`** — `FullBuild` dispatches Go to `loadGoModule` when registered as `PackageLoader`; strips Go files from the per-file fanout on success; on `errNoGoModule`, falls back to a freshly constructed `*GoParser` for Go. `parserVersionTag` now encodes the backend (`go=native` vs `go=treesitter`), AND `Build()` now compares **both** `SchemaVersion` and `ParserVersion` so flipping the flag invalidates caches cleanly.
- **`internal/graph/cache/incremental.go`** — when any `*.go` / `go.mod` / `go.sum` change is in the change set AND the Go parser implements `PackageLoader`, re-run `LoadModule` and wipe-and-reinsert all Go-owned rows via new `Store.DeleteByLanguage("go")` (transactional). Non-Go changes stay on the existing per-file path.
- **`internal/graph/store/store.go`** — new `DeleteByLanguage(lang)` (transactional edge + node wipe by language column) and `CountNodes`/`CountEdges` helpers used by parity tests.
- **`internal/graph/resolver/resolver.go`** — `Resolve` fast-path returns `results` unchanged when no `external::*` edges AND no `InterfaceMethods` — the native Go path becomes inert while TS path-alias rewriting + tree-sitter Go path stay on the slow lane.
- **Build-tag-gated tests** (`//go:build parity` and `//go:build bench`) — produce committed markdown reports:
  - [`parity-report.md`](docs/plans/2026-05-20-graph-native-ast-parity-report.md): tree-sitter vs native on devpilot itself. 3354 `external::*` placeholders correctly dropped; 67 heuristic mis-rewrites no longer fire (wins); 118 candidate "misses" — spot-checks attribute most to tree-sitter heuristics that coincidentally hit real native IDs.
  - [`bench.md`](docs/plans/2026-05-20-graph-native-ast-bench.md): native 1.085 s cold (well under 5 s budget); tree-sitter 71 ms; heap 780 MB native vs 4 MB tree-sitter. Design budget honest-updated to ≤ 1 GB.
- **Docs**: `CLAUDE.md` and `README.md` mention the env flag with trade-offs; Phase 5 reframed for Go (coverage check via `gopls workspace/symbol`, not precision/recall gate — since both consume `go/types` the comparison is near-tautological; precision/recall stays for TS/Rust where the LSP is genuinely independent).

## Multi-file fixture

`internal/graph/parser/testdata/go_native/` — a 6-file `example.com/native` module covering: cross-package call (`pkg/b/b.go` calls `pkg/a.Greet`), interface + impl + method-key collision negative case (`Console.Speak` vs `PartialSpeaker.Speak(int)`), test function (`TestGreet`), type alias. `go list ./...` from the repo root does not pick it up (isolation verified).

## Review Guide

**Start with [`docs/plans/2026-05-20-graph-native-ast-design.md`](docs/plans/2026-05-20-graph-native-ast-design.md)** for the why + scope (what Go repo shapes are supported) + tradeoffs.

Then read in this order:

1. **`internal/graph/parser/go_native.go`** — the meat. Three things to watch:
   - **`objIndexKey`** (line ~43) — receiver-aware method keying. `Console.Speak` and `PartialSpeaker.Speak` must NOT collide; tests cover this.
   - **Three-pass structure** in `LoadModule`: pass 1 mints all decls + builds `objIndex` + `inModule`; pass 2 walks call sites for `calls`/`tests`; pass 3 emits `implements` via `types.Implements`. Pass 1 must finish across **all** packages before pass 2 starts because cross-package callees may live in later-walked packages.
   - **`Tests:true` ownership rule** — non-test package wins for shared files; `_test.go` files keep their nodes under their source-side package.

2. **`internal/graph/cache/builder.go`** lines 85–185 — the dispatch. The `useNative` / `nativeErrs` / `fallbackGo` triplet handles three states: native success, non-module fallback, no native parser at all.

3. **`internal/graph/cache/incremental.go`** lines ~55–110 — `goReload` detection and the wipe-and-reinsert path. Honest: any Go change re-types the whole module. File-level Go incremental is impossible with `go/types`; deferred caching of `*packages.Package` between builds is tracked as a follow-up, not in N1.

4. **`internal/graph/resolver/resolver.go`** — the fast-path skip. Confirm it doesn't break the existing tree-sitter rewrite or TS path-alias resolution.

5. **`docs/plans/2026-05-20-graph-native-ast-parity-report.md`** — the empirical view. 68.7% `(src, kind)` coverage looks low at first glance; the report explains why (tree-sitter emits `imports` edges to `external::stdlib` that native correctly drops).

## Trade-offs (called out, not hidden)

- **Build time**: ~15x slower than tree-sitter on devpilot (1.085 s vs 71 ms). Acceptable for a tool that runs per-PR or per-task, not per-keystroke. Larger repos amortise better.
- **Memory**: 780 MB peak heap on devpilot. Design budget revised from 500 MB → ≤ 1 GB; called out in `bench.md`.
- **Incremental Go**: collapses to whole-module retypecheck on any `*.go`/`go.mod`/`go.sum` change. The `*packages.Package` between-build cache is a deferred optimisation, not N1.
- **Risk**: default-off via env flag. Zero behavioural change for current users on this merge.

## Test plan

- [x] `make test` — full suite green (no `parity`/`bench` tags). 16 packages pass.
- [x] `make lint` — golangci-lint clean.
- [x] `go test -tags=parity ./internal/graph/cache/` — parity test passes (log-only by design); report regenerated.
- [x] `go test -tags=bench -bench='Benchmark.*FullBuild' -benchtime=1x -count=3 ./internal/graph/cache/ -run='^$'` — both backends benched; bench markdown current.
- [x] Live smoke: `DEVPILOT_GRAPH_GO_BACKEND=native devpilot graph build` against this repo + `devpilot graph preflight` work end-to-end.
- [x] `DEVPILOT_GRAPH_GO_BACKEND=` (unset) → tree-sitter behaviour unchanged.
- [x] Cache invalidation test: poison `meta.ParserVersion` → `Build()` triggers `FullBuild` (covered by `TestParserVersionMismatchRebuilds`).

## Follow-ups (not blocking; queued for a small cleanup PR)

1. `BuildIncremental.FilesParsed` miscount under native dispatch (cosmetic in CLI output).
2. Inline `HasPrefix + TrimPrefix` at three hot sites in `go_native.go` — replace with the existing `relFromPrefix` helper.
3. `parserVersionTag` literal still says `phase2:` — change to `phase-n1:` or drop the phase token entirely.
4. `ParseError.Path` falls back to `pkg.PkgPath` (not a file path) when error position is empty — consider a sentinel.
5. Add `go.work` block-form (`use (./a ./b)`) fixture.
6. Update parent design doc (`2026-05-19-devpilot-graph-design.md`) where the original 500 MB heap budget is still quoted.

## For reviewers (human)

- [ ] Sanity-check the parity report's "Reason C" sample (potential native misses) — confirm they're tree-sitter heuristic coincidences, not real native bugs.
- [ ] Confirm the env-flag default-off is right for a one-release bake period (alternative: opt-in via PR-specific override).
- [ ] Decide whether N1.17 (delete tree-sitter Go) should ride a separate PR or be queued with the follow-ups above.


SiyuQian (commented):
## Verdict: Request changes

One Critical correctness gap (D-1) and a handful of Important findings around the build-time error / fallback surface. The core design (three-pass walk, receiver-aware keying, env-flag gating) is solid; what needs work is the error envelope around `loadGoModule`.

## Strengths

- **Architecture seam is clean.** `PackageLoader` is a small additive interface; `cache.Builder` dispatches via type assertion with a deterministic merge (sorted keys, native-before-fanout). Non-Go languages untouched.
- **Receiver-aware method keying is correct.** `objIndexKey` strips pointer layers and keys methods via `*types.Named.Obj().Name()`, so `Stack[T]` vs `Queue[T]` cannot collide. Regression test included.
- **`Tests:true` ownership rule is the right one** — “non-`.test` first, sorted by `pkg.ID`” preserves both source-package nodes for shared files and test-only nodes for `_test.go` files.
- **Cache-flip safety is enforced.** `parserVersionTag` encodes the backend AND `Build()` compares both `SchemaVersion` and `ParserVersion`; regression test (`TestParserVersionMismatchRebuilds`) covers it.
- **Default-off via env flag** — zero behavioural change on merge.

## TL;DR

Native backend works end-to-end on the happy path (single-module repos). The fallback paths and error envelopes are where the regressions hide:

1. **D-1 (Critical)** — non-module repo + incremental rebuild + native flag silently loses Go nodes (real data loss). Mirror FullBuild's `fallbackGo` pattern in `BuildIncremental`.
2. **A-1, A-2, A-3, A-4, B-1** — hard-fail / silent-incomplete / lock-hold / env-variability / go.work-not-tracked surfaces the happy-path fixture doesn't exercise.
3. **A-5** — external test packages (`package foo_test`) aren't in the fixture; potential `objIndex` collision when both `foo` and `foo_test` declare the same simple name.

## Sweep (Agent A blind-spot questions)

1. **Large real-world repo?** Build-lock-held-during-LoadModule (A-2) and env-variability (A-3) surface here.
2. **Concurrency under native dispatch?** Non-Go fanout still parallel, but the lock-hold window grows from seconds to minutes (A-2).
3. **Failure isolation under partial type errors?** Silent incompletion — `pkg.IllTyped` packages emit nodes but lose edges, and `ParseError`s never reach the build summary (A-4).
4. **Determinism under env variation?** No — `parserVersionTag` doesn't encode GOOS / GOARCH / CGO_ENABLED / build tags (A-3).
5. **External test package handling?** Untested; potential collision (A-5).

## Counts

- Critical: 1 (D-1)
- Important: 7 (A-1, A-2, A-3, A-4, A-5, B-1, E-1)
- Minor: 3 (B-4, B-5, E-3)
- Nits: 0

## Open Questions for the author

1. **Partial-type-check semantics**: should `pkg.IllTyped` packages emit edges-where-possible (current) or be hard-failed? The answer determines whether A-4 is “fix now” or “document as known limitation.”
2. **Lock-hold trade in A-2**: acceptable for v1? If yes, document it and bump the timeout when native is selected.
3. **N1.17 (delete tree-sitter Go) sequencing**: after the env flag flips by default, or kept separate from the rollout PR?

internal/graph/cache/incremental.go
```diff
@@ -48,18 +50,66 @@ func (b *Builder) BuildIncremental(prev Meta) (BuildResult, error) {
...
+	useNative := false
+	if goReload {
+		if goP := b.reg.ForLanguage("go"); goP != nil {
+			if loader, ok := goP.(parser.PackageLoader); ok {
+				res, lerr := loadGoModule(loader, b.repo)
+				switch {
+				case lerr == nil:
+					useNative = true
+					nativeResults = res
+				case errors.Is(lerr, errNoGoModule):
```
SiyuQian, line 86: **Severity: Critical | Confidence: 95**

When `loadGoModule` returns `errNoGoModule` here, `useNative` stays false but `goReload` stays true — and there's no `fallbackGo` setup like `FullBuild` has. Go files in `toParse` then route through `b.reg.ForPath(p)` which returns `*GoNativeParser`; its `Parse` is a no-op that returns `(ParseResult{}, nil)`. Result: **incremental rebuilds in non-module repos with the flag set silently produce zero nodes for changed Go files** (real data loss).

Mirror the FullBuild pattern: capture a `goFallback *parser.GoParser` in this branch and apply it in the per-file loop at lines 114–130 when `filepath.Ext(p) == ".go"`. A regression test covering non-module + native + incremental would catch this.

internal/graph/cache/builder.go
```diff
@@ -80,6 +82,49 @@ func (b *Builder) FullBuild() (BuildResult, error) {
...
+				nativeResults = res
+			case errors.Is(lerr, errNoGoModule):
+				// Non-module repo: record a synthetic ParseError so the signal
+				// surfaces in results, but keep going via the tree-sitter fanout.
+				nativeErrs = append(nativeErrs, parser.ParseError{
+					Path:    "",
+					Message: "native Go backend skipped: " + lerr.Error(),
+				})
+			default:
+				return BuildResult{}, fmt.Errorf("native Go load: %w", lerr)
```
SiyuQian, line 122: **Severity: Important | Confidence: 90**

Any non-`errNoGoModule` error from `loadGoModule` (e.g. transient `packages.Load` failure, vendored file with a syntax error, `parse go.work` failure, or `LoadModule`'s own “no usable packages loaded”) aborts the entire build — including all TypeScript / JavaScript / Rust files. The tree-sitter path would have degraded gracefully per-file.

Consider treating non-`errNoGoModule` errors as a soft fallback: append a `ParseError`, set `fallbackGo = parser.NewGoParser()`, and continue. Hard-failing the whole multi-language build because Go failed is a regression vs. the previous behaviour.

internal/graph/cache/builder.go
```diff
@@ -80,6 +82,49 @@ func (b *Builder) FullBuild() (BuildResult, error) {
...
+	// Native Go backend dispatch: when the registered Go parser implements
+	// PackageLoader (the native backend), call LoadModule once per module and
+	// strip Go files from the per-file fanout. On non-module repos we silently
+	// fall back to per-file Parse so the build doesn't crash on stray inputs.
+	var nativeResults map[string]parser.ParseResult
+	var nativeErrs []parser.ParseError
+	useNative := false
+	if goP := b.reg.ForLanguage("go"); goP != nil {
+		if loader, ok := goP.(parser.PackageLoader); ok {
+			res, lerr := loadGoModule(loader, b.repo)
```
SiyuQian, line 107: **Severity: Important | Confidence: 85**

The build lock is acquired at line ~62 of this same function (pre-existing) and `loadGoModule` runs inside it here. On a Kubernetes-sized module the type-check takes minutes; any concurrent `devpilot graph build/preflight` against the same repo will trip the 60s `AcquireBuildLock` timeout — e.g. the PR-review skill running preflight while a user-triggered build is in flight.

Suggest moving the `loadGoModule` call BEFORE `AcquireBuildLock` — only the SQLite writes need exclusivity. As a smaller mitigation, bump the lock timeout when the native backend is selected and document the lock-hold expectation.

internal/graph/cache/builder.go
```diff
@@ -186,5 +254,14 @@ func (b *Builder) Build() (BuildResult, error) {
 }
 
 func parserVersionTag(reg *parser.Registry) string {
-	return "phase2:" + strings.Join(reg.Languages(), ",")
+	langs := reg.Languages()
+	parts := make([]string, len(langs))
+	for i, lang := range langs {
+		if lang == "go" {
+			parts[i] = "go=" + reg.GoBackend()
```
SiyuQian, line 276: **Severity: Important | Confidence: 88**

`parserVersionTag` encodes only the backend selector. `packages.Load` honors ambient `GOOS` / `GOARCH` / `GOFLAGS` (build tags) / `CGO_ENABLED`: a graph built on macOS includes `*_darwin.go` files; the same repo built on a Linux CI agent sees `*_linux.go`. Both runs share the same `ParserVersion`, so `Build()` happily takes the incremental path on top of a stale-for-this-toolchain graph (no `.go` file changed → `goReload` stays false → no rebuild).

When the native backend is active, fold `runtime.GOOS`, `runtime.GOARCH`, and `os.Getenv("CGO_ENABLED")` / `GOFLAGS` into the tag. Snapshotting `go env -json` is overkill; those four selectors cover the practical drift.

internal/graph/parser/go_native.go
```diff
@@ -0,0 +1,683 @@
...
+			}
+		}
+		// Skip the synthetic test binary; its files are owned by the source/_test pkgs.
+		if ent.isDotTest {
+			continue
+		}
+		if pk.Fset == nil || len(pk.Syntax) == 0 {
+			continue
+		}
+		usable++
```
SiyuQian, line 299: **Severity: Important | Confidence: 82**

`usable++` fires for any package with non-empty `Syntax`, regardless of `pkg.IllTyped` or non-empty `pkg.Errors`. When type info is incomplete (one undeclared import poisons a whole subgraph), pass 2's `TypesInfo.Uses[ident]` returns `nil` for many call sites and the walker silently returns true. The build “succeeds” but emits a graph missing edges for every affected symbol.

The synthetic `ParseError`s under the `""` key aren't surfaced upstream — `builder.go` aggregates only `Nodes` and `Edges`. Two clean options:

1. Wire `r.Errors` into `BuildResult` (a `Warnings []string` field) so the CLI can show them.
2. Treat `pkg.IllTyped == true` as a hard error and short-circuit before pass 2.

Current state is the worst of both — looks green but loses information silently.

internal/graph/parser/go_native.go
```diff
@@ -0,0 +1,683 @@
...
+
+	// Process non-test-binary packages first to ensure source-package ownership
+	// wins over the `pkg.test` synthetic variant when a file appears in both.
+	type pkgEntry struct {
+		pkg       *packages.Package
+		isDotTest bool
+	}
+	entries := make([]pkgEntry, 0, len(pkgs))
+	for _, pk := range pkgs {
+		entries = append(entries, pkgEntry{pkg: pk, isDotTest: strings.HasSuffix(pk.ID, ".test")})
```
SiyuQian, line 191: **Severity: Important | Confidence: 75**

The `.test` synthetic binary is filtered here, but `packages.Load(Tests:true)` also synthesises an external test package whose ID looks like `foo_test [foo.test]` — doesn't end in `.test`. The fixture under `testdata/go_native/` uses `package a` for both `a.go` and `a_test.go`, so external test packages (`package foo_test`) aren't exercised at all.

Failure mode: if `foo` and `foo_test` both declare a function with the same simple name (legal Go), `objIndex[pkg.Path()::Name]` keeps only one and calls from external-test code may resolve to the wrong target or drop. Add a fixture with `package foo_test`, assert distinct node IDs survive, and tighten the filter rule if needed.

internal/graph/cache/incremental.go
```diff
@@ -48,18 +50,66 @@ func (b *Builder) BuildIncremental(prev Meta) (BuildResult, error) {
...
 
+	// Native Go has no file-level incremental path: go/types needs the whole
+	// module. If a Go source file (or go.mod/go.sum) changed AND the registered
+	// Go parser is a PackageLoader, re-run LoadModule for the entire module,
+	// drop every Go-owned row, and inject the fresh native results into the
+	// merge. Non-Go files still take the existing per-file path.
+	goReload := false
+	for _, list := range [][]string{changed.Added, changed.Modified, changed.Deleted} {
+		for _, p := range list {
+			if filepath.Ext(p) == ".go" || p == "go.mod" || p == "go.sum" {
```
SiyuQian: **Severity: Important | Confidence: 88**

`go.work` is missing from the `goReload` predicate, but `loadGoModule` consumes it to decide which modules to scan. Editing `go.work` (add/remove a `use` directive) without touching any `.go` file leaves the graph stale — newly added modules never enter the graph; removed modules' rows never wiped.

Add `|| p == "go.work" || p == "go.work.sum"` to the predicate.

internal/graph/cache/builder.go
```diff
@@ -80,6 +82,49 @@ func (b *Builder) FullBuild() (BuildResult, error) {
...
+		if loader, ok := goP.(parser.PackageLoader); ok {
+			res, lerr := loadGoModule(loader, b.repo)
+			switch {
+			case lerr == nil:
+				useNative = true
+				nativeResults = res
+			case errors.Is(lerr, errNoGoModule):
+				// Non-module repo: record a synthetic ParseError so the signal
+				// surfaces in results, but keep going via the tree-sitter fanout.
+				nativeErrs = append(nativeErrs, parser.ParseError{
```
SiyuQian, line 117: **Severity: Minor | Confidence: 75**

The synthetic `ParseError` for the `errNoGoModule` fallback uses `Path: ""`, which collides with `go_native.go`'s own synthetic errors key (also `""`). Downstream consumers grouping by `Path` see two semantically distinct “empty-path” sources merged.

Use a distinct sentinel like `"<native-go:non-module>"` so the signal is unambiguous if anyone surfaces it (per A-4, today nobody does — separate fix).

internal/graph/cache/native_go.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			usePaths = append(usePaths, abs)
+		}
+		sort.Strings(usePaths)
+
+		merged := map[string]parser.ParseResult{}
+		for _, up := range usePaths {
+			sub, lerr := loader.LoadModule(up)
+			if lerr != nil {
+				return nil, fmt.Errorf("load module %s: %w", up, lerr)
```
SiyuQian: **Severity: Minor | Confidence: 80**

If any single `use` directory in `go.work` returns `“no usable packages loaded”` (e.g. a placeholder module with `go.mod` but zero `.go` files), this hard-fails the entire workspace build. A permissive workspace should tolerate that.

Treat the “no usable packages” error from a single use-module as a `ParseError` and continue with the rest. Alternatively, change `GoNativeParser.LoadModule` to return a softer signal (`(emptyMap, nil)`) when the module is empty.

internal/graph/parser/go_native.go
```diff
@@ -0,0 +1,683 @@
...
+		return false
+	}
+	tyPkg := tyObj.Pkg()
+	if tyPkg == nil {
+		return false
+	}
+	return tyPkg.Path() == "testing" && tyObj.Name() == "T"
+}
+
+// LoadModule type-checks the whole Go module rooted at repoRoot and returns a
```
SiyuQian, line 145: **Severity: Minor | Confidence: 95**

Doc-comment drift. The comment says results contain “function and method nodes” with “contains edges from file → declaration.” The implementation also emits `struct` / `interface` / `type` nodes and four additional edge kinds (`imports` / `calls` / `tests` / `implements`).

List everything `LoadModule` produces so the next reader doesn't need to trace 500 lines to find out.

internal/graph/parser/go_native.go
```diff
@@ -0,0 +1,683 @@
...
+
+func (p *GoNativeParser) Language() string {
+	return "go"
+}
+
+func (p *GoNativeParser) Extensions() []string {
+	return []string{".go"}
+}
+
+// Parse is intentionally a no-op; the native backend produces results via
```
SiyuQian, line 34: **Severity: Nit | Confidence: 90**

This correctly flags `Parse` as a no-op, but `LoadModule`'s doc comment doesn't contrast with the tree-sitter `GoParser`. A one-liner here (or atop `LoadModule`) that “unlike the tree-sitter backend (single-pass per-file, post-resolver), the native backend produces fully-resolved edges in a single whole-module pass” would save the next reader a 30-minute spelunk.

internal/graph/cache/incremental.go
SiyuQian: Fixed in 885c647. Added fallback `*parser.GoParser` in the `errNoGoModule` branch and routed Go files through it in the per-file loop. Regression test `TestIncrementalNativeGoNonModuleFallback` asserts a new `Farewell` function in a non-module repo lands in the graph after an incremental rebuild.

internal/graph/cache/builder.go
SiyuQian: Deferring this one. A soft fallback on every `LoadModule` error means a corrupt `go.mod` or transient `packages.Load` failure silently produces a half-graph with no signal. Hard-fail is the right default for v1; better to make the failure loud and document the failure surface. Queued as N1.18 along with surfacing `pkg.IllTyped` (A-4) so the user gets a real warning channel instead of a silent fallback.

internal/graph/cache/builder.go
SiyuQian: Bumped the lock timeout from 60s to 5min when the native backend is selected (5d956df). Won't move `loadGoModule` outside the lock — two concurrent native builders parsing simultaneously then one wiping the other's writes is a worse failure mode.

internal/graph/cache/builder.go
SiyuQian: Fixed in 5d956df. `parserVersionTag` now appends an `env=goos=…,goarch=…,cgo=…,goflags=…` signature when the native backend is selected; tree-sitter is unaffected. Regression test `TestParserVersionTagEncodesGoEnv` covers CGO_ENABLED + GOFLAGS flips.

internal/graph/parser/go_native.go
SiyuQian: Deferring. Surfacing `pkg.IllTyped` needs a new `BuildResult.Warnings` field and CLI envelope plumbing — out of scope for the cleanup PR. Tracked as N1.18 alongside A-1 so they ship together. Current behavior (silent partial graph) is the worst of both; agreed it needs fixing soon.

internal/graph/parser/go_native.go
SiyuQian: Deferring. Need to add a `package foo_test` external-test-package fixture and decide whether the filter rule or the objIndex key needs to tighten. Queued as N1.18 with a dedicated fixture so we can verify the collision scenario before changing logic.

internal/graph/cache/incremental.go
SiyuQian: Fixed in 885c647. Predicate now also triggers on `go.work` and `go.work.sum` changes.

internal/graph/cache/builder.go
SiyuQian: Fixed in 5d956df (via the `nonModuleErrorPath` sentinel introduced in b536d7d). The non-module fallback now uses `<native-go:non-module>` as the Path, distinct from go_native.go's per-package `""` key.

internal/graph/cache/native_go.go
SiyuQian: Fixed in b536d7d. Per-use-module `LoadModule` errors now land as `ParseError` entries under the synthetic `""` key in the merged result and the loop continues with remaining modules. A workspace with one empty placeholder module no longer aborts the full build.

internal/graph/parser/go_native.go
SiyuQian: Fixed in 00f6519. Doc comment now lists all node kinds (file, function, method, struct, interface, type) and all edge kinds (contains, imports, calls, tests, implements) that LoadModule actually emits.

internal/graph/parser/go_native.go
SiyuQian: Fixed in 00f6519. `Parse`'s doc comment now contrasts the native backend's single-pass whole-module shape with the tree-sitter backend's per-file + post-resolver shape.

PR #115, opened 2026-05-21
fix(release): native CGO matrix build for tree-sitter binaries

SiyuQian:
## Summary
- v0.18.1 release failed because release.yml cross-compiles with default `CGO_ENABLED=0`, but `internal/graph/parser/{typescript,javascript,rust}.go` still depend on `smacker/go-tree-sitter`, whose `iter.go` references CGO-only `Node` types (`undefined: Node`). Background: the Go parser was migrated to a native AST backend (#112, #113) but TS/JS/Rust still need tree-sitter.
- Switch release to a native build matrix: `macos-14` → darwin/arm64, `macos-13` → darwin/amd64, `ubuntu-latest` → linux/amd64. Each builds with `CGO_ENABLED=1`, uploads its artifact, and a `publish` job aggregates, checksums, and uploads to the release.
- Logged the tree-sitter CGO coupling in `docs/exec-plans/tech-debt-tracker.md` — revisit going CGO-free once a maintained wazero/WASM tree-sitter binding (e.g. `malivvan/tree-sitter`) ships v1, or `microsoft/typescript-go` exposes a stable AST.

## Test plan
- [ ] Cut a fresh draft release tag and confirm all three artifacts + `checksums.txt` upload.
- [ ] Sanity-check each downloaded binary runs (`devpilot --version`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,9 +24,9 @@ jobs:
             goarch: amd64
     runs-on: ${{ matrix.runner }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
 
@@ -40,7 +40,7 @@ jobs:
           OUT=devpilot-${{ matrix.goos }}-${{ matrix.goarch }}
           go build -ldflags "-X main.version=${VERSION}" -o "${OUT}" ./cmd/devpilot
 
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v7
         with:
           name: devpilot-${{ matrix.goos }}-${{ matrix.goarch }}
           path: devpilot-${{ matrix.goos }}-${{ matrix.goarch }}
@@ -50,7 +50,7 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/download-artifact@v4
+      - uses: actions/download-artifact@v8
         with:
           path: dist
           merge-multiple: true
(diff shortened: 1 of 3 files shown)


SiyuQian (commented):
**Verdict: ship it.**

## TL;DR
Matrix-CGO release is the right fix for the v0.18.1 break — same grammar, same correctness, no parser rewrite. Action bumps are clean for our usage.

## Strengths
- Aggregating into a single `publish` job keeps checksums/uploads atomic instead of racing per-matrix.
- `if-no-files-found: error` fails loud if a build produces nothing.
- Tech-debt entry captures the right escape hatch (wazero+WASM tree-sitter, microsoft/typescript-go) without committing us to do it now.

## Unknown-Unknowns Sweep
- **Action breaking changes (v4→v6/v7/v8 jumps)**: verified against release notes — all bumps are Node 24 runtime + minor internal changes; none affects our inputs (`merge-multiple`, `version`, `files`, `token` all unchanged). upload-v7 ↔ download-v8 cross-compatible via the shared `@actions/artifact` v4+ backend.
- **macos-13 runner**: still GA, but on GitHub's deprecation track for a future quarter. When it goes, darwin/amd64 will need cross-compile from macos-14 ARM with `-arch x86_64` + matching CGO flags. Not blocking; worth noting for the next 6–12 months.
- **`gh release upload` without `--clobber`**: will fail on re-run if assets already exist. Acceptable for first-attempt releases; can add `--clobber` if we start re-running this workflow on failed publishes.
- **linux binary now dynamically linked to glibc** (CGO_ENABLED=1). Fine for Debian/Ubuntu/Fedora users; Alpine/musl users would need a separate static build path — not in scope.

## Counts
- Inline findings (≥70 confidence): 0
- Open questions: 0

## Open Questions
None.

PR #132, opened 2026-05-22
fix(graph/parser): nil-guard TypesInfo in LoadModule TYPE-spec branch

SiyuQian:
## Summary
- The func-decl branch in `LoadModule` (line ~399) correctly guards `if pk.TypesInfo != nil`, but the TYPE-spec branch below did not, so `packages.Load` returning a package whose `TypesInfo` is nil (e.g. failed type-check) panics.
- Adds matching `if pk.TypesInfo == nil { continue }` guard with a comment cross-referencing the func-decl branch.

Closes #124

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/graph/parser/...` (37 pass)
- [x] `make lint`
- [ ] No targeted regression test added — constructing a `*packages.Package` with nil TypesInfo + TypeSpecs deterministically is brittle; guard mirrors an already-tested pattern.

### DIFF
--- a/internal/graph/parser/go_native.go
+++ b/internal/graph/parser/go_native.go
@@ -420,6 +420,12 @@ func (p *GoNativeParser) LoadModule(repoRoot string) (map[string]ParseResult, er
 				if !ok || gd.Tok != token.TYPE {
 					continue
 				}
+				// Mirror the func-decl branch's guard above: when packages.Load
+				// returns a package that failed to type-check, TypesInfo can be
+				// nil and dereferencing it panics. Skip the type-spec block.
+				if pk.TypesInfo == nil {
+					continue
+				}
 				for _, spec := range gd.Specs {
 					ts, ok := spec.(*ast.TypeSpec)
 					if !ok || ts.Name == nil {


SiyuQian (commented):
## Verdict: approve

Minimal nil-guard that mirrors the existing func-decl branch pattern at line 399 and matches the pass-2 guard at line 517. Prevents a panic when `packages.Load` returns a package whose type-checking failed but still produced AST decls. Low blast radius, correct control flow.

### Graph check
`devpilot graph preflight` confirms `LoadModule` has 7 test callers (all in `go_native_test.go`) and is not a hub; no cross-community edges or interface-shape changes. All three TypesInfo dereferences in LoadModule (lines 400, 435, 517) are now guarded.

### Strengths
- Guard pattern uniform with existing func-decl guard (L399) and pass-2 guard (L517).
- `continue` (not `break`) is the right control-flow: loop iterates per-decl within a file, other decls remain processable.
- Comment cites the sibling guard — self-documenting.
- Skipping the test is defensible: reproducing a `packages.Package` with nil TypesInfo + non-nil Syntax requires either a real broken-module fixture or unexported-field reflection.

PR #133, opened 2026-05-22
fix(graph/resolver): handle /* */ block comments in stripJSONComments

SiyuQian:
## Summary
- `stripJSONComments` only handled `//` line comments. A `/* ... // ... */` block comment in a real tsconfig.json triggered the line-comment branch on the embedded `//`, eating through the next newline including `*/` and corrupting the JSON.
- Extends the state machine outside string literals to also strip `/* ... */` block comments (unterminated blocks consumed to EOF).
- Adds table-driven tests covering block comment with embedded `//`, block comment inside string literal (must survive), unterminated block, line comments, and the existing URL-in-string case.

Closes #125

## Test plan
- [x] `go test ./internal/graph/resolver/...` (16 pass incl. new cases)
- [x] `go build ./...`
- [x] `make lint`

### DIFF
--- a/internal/graph/resolver/tsconfig.go
+++ b/internal/graph/resolver/tsconfig.go
@@ -131,10 +131,9 @@ func (r *TSConfigResolver) resolve(spec string) (string, bool) {
 	return "", false
 }
 
-// stripJSONComments removes JSONC `//` line comments while preserving
-// occurrences of `//` inside string literals (so URLs and path values
-// like "https://x" survive intact). Block comments are not handled —
-// add when a real tsconfig in the wild needs them.
+// stripJSONComments removes JSONC `//` line comments and `/* ... */` block
+// comments while preserving occurrences of `//` or `/*` inside string literals
+// (so URLs and path values like "https://x" or "http://x/*foo*/" survive intact).
 func stripJSONComments(b []byte) []byte {
 	out := make([]byte, 0, len(b))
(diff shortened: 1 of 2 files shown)


SiyuQian (commented):
## Verdict: approve

Correctly extends `stripJSONComments` to handle JSONC block comments with a clean state machine that preserves string-literal contents and safely terminates on unterminated comments. Reachability into `NewTSConfigResolver` (called by graph cache `FullBuild`/`BuildIncremental`) confirmed via codegraph, so the fix actually exercises in production tsconfig parsing.

### Graph check
`devpilot graph preflight` shows `stripJSONComments` → `loadTSConfig` → `NewTSConfigResolver` with 3 external callers including both full and incremental graph builds. New test `TestStripJSONCommentsParsesRealisticTSConfig` covers the regression from the issue.

### Strengths
- State machine ordering correct: `inString` check runs before comment checks, so `/*` and `//` inside string literals are preserved.
- Unterminated block comment is bounded — when inner scan exits with `i+1 >= len(b)`, `i` is explicitly set to `len(b)`, so the outer loop can't infinite-loop or read OOB.
- Terminator detection at `/` after `*` handled correctly by single `i++` inside the branch plus outer loop `i++`.
- Tests cover bug-from-issue scenario, string-embedded markers, multi-line spans, escaped quotes, EOF termination.

PR #134, opened 2026-05-22
test(graph): cover cli_query dispatcher and JSON payload shape

SiyuQian:
## Summary
Adds `internal/graph/cli_query_test.go` covering the `devpilot graph query` dispatcher and the JSON payload shape (the public contract consumed by skills/agents).

- Dispatcher patterns: callers_of, callees_of, tests_for, implementors_of, hubs, context — each asserts envelope shape and exact field names.
- Error envelopes: unknown_pattern, cache_missing, repo_invalid — non-zero rc with structured error code.
- Nil-normalization: empty result lists emit `[]` not `null`.
- Fixture seeds a `store.Store` at `cache.GraphDB(home, key)` with a topology covering call/test/implements/hub edges plus an on-disk source file for the context snippet.

Closes #126

## Test plan
- [x] `go test ./internal/graph/...` (149 tests pass)
- [x] `make lint`

### DIFF
--- a/internal/graph/cli_query_test.go
+++ b/internal/graph/cli_query_test.go
@@ -1,0 +1,336 @@
+package graph
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/siyuqian/devpilot/internal/graph/cache"
+	"github.com/siyuqian/devpilot/internal/graph/store"
+)
+
+// seedQueryStore creates a fresh DEVPILOT_HOME, a repo dir, and a seeded
+// graph.db at the cache path that runQuery will resolve.
+//
+// Fixture topology:
+//
+//	caller  -> target   (calls)
+//	target  -> leaf     (calls)
+//	tester  -> target   (tests)
+//	impl    -> iface    (implements)
+//	hub_a   -> hub      (calls)
+//	hub_b   -> hub      (calls)
+//	hub_c   -> hub      (calls)
+//
+// target/caller live in src.go with known line ranges so context can read them.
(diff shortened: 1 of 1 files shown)


SiyuQian (commented):
## Verdict: approve

Test-only PR adds in-process coverage for the `runQuery` dispatcher across all six patterns plus three error envelopes and nil-to-empty-array normalization. Field names asserted match `runQuery`'s actual output exactly. Fixture uses real `cache.RepoKey`/`EnsureDirs`/`GraphDB` paths and `store.Open`, so it mirrors production layout.

### Graph check
Verified against `c951d89d:internal/graph/cli_query.go` and `cache/paths.go`. Test exercises `runQuery` directly in-process, seeds SQLite at `cache.GraphDB(home, key)` under `t.TempDir()`+`DEVPILOT_HOME`, decodes the envelope from `emit/envelope.New`. Every asserted key (`pattern_result.callers/callees/tests/implementors/hubs/context`, snippet keys, error.code values) is the exact key the production code emits.

### Strengths
- Calls `runQuery` in-process via `captureStdout` rather than shelling to the binary — correct unit-test boundary.
- Uses the real cache helpers instead of hardcoded paths, so a future cache layout change won't silently desync the test.
- Pins `caller_count` as the canonical hub key (comment notes skills depend on it) — locks the public JSON contract.
- `TestRunQueryEmptyResultIsArray` pins the `nil → []` normalization that schema oneOf branches rely on.
- Covers all three error envelopes (unknown_pattern, cache_missing, repo_invalid) including the rc!=0 contract.

PR #135, opened 2026-05-22
test(graph/resolver): cover addImplementsEdges and isSuperset

SiyuQian:
## Summary
Adds `internal/graph/resolver/implements_test.go` covering the implements-edge resolver, which underpins the `implementors` query and scanner reachability evidence.

Cases:
- Exact method-set match → edge emitted.
- Missing one method → no edge.
- Superset (struct has extra methods) → edge emitted.
- Empty `interface{}` → no implements edges for any struct (no match-everything).
- Unrelated struct without methods → no false positive on same-name.
- Cross-file methods within a package (struct in `a.go`, method in `b.go`) → no edge (locks in v1 limitation).

Also adds a direct table test for `isSuperset`.

Closes #127

## Test plan
- [x] `go test ./internal/graph/resolver/...` (20 pass)
- [x] `make lint`

### DIFF
--- a/internal/graph/resolver/implements_test.go
+++ b/internal/graph/resolver/implements_test.go
@@ -1,0 +1,224 @@
+package resolver
+
+import (
+	"testing"
+
+	"github.com/siyuqian/devpilot/internal/graph/parser"
+	"github.com/siyuqian/devpilot/internal/graph/store"
+)
+
+// hasImplementsEdge reports whether any result in the batch contains an
+// implements edge from src to dst.
+func hasImplementsEdge(results []parser.ParseResult, src, dst string) bool {
+	for _, r := range results {
+		for _, e := range r.Edges {
+			if e.Kind == "implements" && e.Src == src && e.Dst == dst {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// countImplementsEdges counts every implements edge across the batch.
+func countImplementsEdges(results []parser.ParseResult) int {
+	n := 0
+	for _, r := range results {
+		for _, e := range r.Edges {
+			if e.Kind == "implements" {
+				n++
+			}
+		}
+	}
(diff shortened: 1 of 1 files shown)


SiyuQian (commented):
## Verdict: approve

Test-only PR adding table-driven coverage for `addImplementsEdges` and `isSuperset`. Fixtures faithfully mirror the Go-native parser's `ParseResult` shape (struct/interface/method Kinds, `<path>::<name>` and `<path>::<container>.<name>` ID conventions, `Container`=receiver type name). Tests drive the real resolver function (no mock/shadow), assertions correctly inspect `r.Edges` for `Kind=="implements"` with Src/Dst, all 14 subtests pass.

### Graph check
`devpilot graph preflight`: only `internal/graph/resolver/implements_test.go` added; no callers, no hubs touched. Resolver verified against parser via `internal/graph/parser/go_native.go` — Kind strings (`method`,`interface`,`struct`) and ID/Container semantics in fixtures match parser output exactly. Cross-file negative case is genuine: `addImplementsEdges` keys structs map by `n.ID` (`a.go::Console`) and matches methods by `n.Path+"::"+n.Container` (`b.go::Console`), so the lookup truly misses — test would fail loudly if the v1 limitation were relaxed.

### Strengths
- Test fixtures use the exact ID conventions emitted by the Go-native parser — actually exercises production attribution logic, not a parallel reality.
- Cross-file v1-limitation case constructed as two ParseResults with distinct Paths, the only configuration that truly demonstrates the `<path>::<container>` exact-match constraint.
- `empty_interface` case guards the documented `len(methodNames) == 0` skip that prevents `interface{}` from matching every concrete type.
- Direct `isSuperset` table test covers equal, strict superset, missing key, empty sub, both empty, disjoint.

PR #157, opened 2026-07-09
fix(pr-creator): make skill safe to run concurrently

SiyuQian:
## What & why

The `devpilot-pr-creator` skill couldn't be run concurrently. When a parent skill (`devpilot-resolve-issues`, `devpilot-auto-feature`) fans out parallel agents that each open a PR, the runs corrupted each other because they:

1. **Mutated shared repo state in one checkout** — `git checkout -b`, staging, and committing all move the shared `HEAD`/index, so parallel runs in the same working directory clobber each other's branch and staged files.
2. **Derived colliding branch names** — deterministic naming makes two parallel runs on similar diffs pick the *same* branch, colliding on push.

This PR is docs-only (the skill is a Markdown instruction file); it adds explicit concurrency rules so overlapping invocations stay isolated.

## Changes

- **New "Concurrency safety" section** in `SKILL.md`:
  - Never mutate a working tree you don't own; when on `main`/`master` in a shared checkout, isolate branch creation in a `git worktree add` instead of `git checkout -b` in place.
  - When already handed a dedicated branch/worktree (the common autonomous case), commit in place — no extra worktree.
  - Let the non-fast-forward push *rejection* arbitrate the race rather than a pre-check; never force-push to resolve it.
  - Clean up any worktree the run created once the PR is open.
- **Branch naming:** append a unique suffix (parent-supplied issue/task number, else short HEAD SHA) in concurrent/autonomous mode so parallel runs don't collide.
- **Auto-recover-on-main:** cross-references the new isolation rule at the `git checkout -b` step.
- Synced the installed `.claude/skills/devpilot-pr-creator/SKILL.md` copy (it was also stale versus `skills/`, so its diff is larger).

## Review Guide

- Start with `skills/devpilot-pr-creator/SKILL.md` — the new **Concurrency safety** section is the substance; everything else cross-references it.
- The `.claude/skills/...` copy is a byte-identical mirror of `skills/...` (verified with `diff`); its larger diff is because it was previously out of sync, not a separate change.
- No code, no build/lint surface touched — this is instruction text only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bgzqdfd5qpH1DM3SVLqsmq)_

### DIFF
--- a/.claude/skills/devpilot-pr-creator/SKILL.md
+++ b/.claude/skills/devpilot-pr-creator/SKILL.md
@@ -14,6 +14,8 @@ license: Complete terms in LICENSE.txt
 **Core principle:** Read the actual diff before writing anything. Every sentence in the description
 must come from code you read, not from branch names or assumptions.
 
+**Operating mode: automatic by default.** Do not stop to confirm routine choices (branch name, commit message, draft body) — derive them from the diff and execute. Only stop for the destructive / ambiguous cases listed under [Hard Stops](#hard-stops). Showing a draft "for approval" before creating is **not** a hard stop; the user can `gh pr edit` after the fact.
+
 ## Quick Reference
 
 | Action | GitHub | GitLab |
@@ -25,29 +27,102 @@ must come from code you read, not from branch names or assumptions.
 | Draft | `--draft` | `--draft` |
 | Push | `git push -u origin HEAD` | `git push -u origin HEAD` |
 
+## Worktree / autonomous invocation
+
+Detect both up front — they change preflight rules.
+
+```bash
+git rev-parse --git-dir            # contains "/worktrees/" → you are in a linked worktree
+git worktree list                  # confirms which worktree is which
+```
+
+**You are in autonomous mode** when this skill was invoked by a parent skill (e.g. `devpilot-resolve-issues`, `devpilot-auto-feature`) rather than directly by a human. In autonomous mode:
+- The "stop and ask the user" gates below become "return control to the parent with the question" — never deadlock waiting for human input that isn't coming.
+- Replace "show the draft to the user before creating" with "include the draft in your final response" so the parent skill can surface it.
+- Do not `cd` out of the worktree; the parent owns cwd.
+
+**Base branch.** Always diff and log against `origin/<default-branch>` (typically `origin/main`), not local `main`. In a worktree the local `main` ref may be stale or absent. Run `git fetch origin <default-branch>` before reading the diff so the comparison is honest.
+
+## Concurrency safety
+
+This skill can run **several times at once** — a parent skill (`devpilot-resolve-issues`, `devpilot-auto-feature`) may fan out parallel agents that each open a PR. Concurrent runs corrupt each other only when they **mutate shared repo state** in the **same working directory**: `git checkout -b`, staging, and committing all move the shared `HEAD`/index, so two runs racing in one checkout clobber each other's branch and staged files. Reads (`git fetch`, `git diff`, `git log`, `gh pr list`) are safe to overlap.
+
+Follow these rules whenever concurrency is possible (autonomous mode, or the user says runs overlap):
+
+- **Never mutate a working tree you do not own.** If you would need `git checkout`/`git checkout -b`/`git switch` to get onto a branch — i.e. you are on `main`/`master` in a shared checkout — do **not** switch in place. Create an isolated worktree instead and do all mutation there:
+  ```bash
+  git fetch origin <default-branch> --quiet
+  git worktree add -b <branch> ../pr-<branch> origin/<default-branch>   # isolated HEAD + index
+  cd ../pr-<branch>                                                     # then stage/commit/push here
+  ```
+  If you were **already handed a dedicated branch/worktree** (the common autonomous case — you are on a feature branch in a linked worktree), you already own it; stage and commit in place, no new worktree needed.
+- **Make the branch name collision-proof.** Deterministic naming (below) makes two parallel runs on similar diffs pick the *same* name and collide on push. In concurrent/autonomous mode, append a short unique suffix to the derived slug — the issue/task number when the parent provides one, otherwise the short HEAD SHA: `<type>/<slug>-<issue-or-sha>`. Never reuse `Date.now()`-style timestamps.
+- **Let push, not a pre-check, arbitrate the race.** Two runs may both pass `git ls-remote` and then both push the same ref. Rely on the non-fast-forward push *rejection* as the signal — if `git push -u origin HEAD` is rejected, re-fetch and inspect per [Hard Stops](#hard-stops); never force-push to resolve it.
+- **Clean up an isolated worktree** you created once the PR is open: `git worktree remove ../pr-<branch>` (skip if it still holds uncommitted work).
+
 ## Preflight Checks
 
 Before anything else, run these in parallel:
 
 ```bash
-git remote get-url origin          # detect platform (github.com → gh, gitlab.com → glab)
-git rev-parse --abbrev-ref HEAD    # current branch
-git status                         # uncommitted changes?
-git log <base>..HEAD --oneline     # commits ahead of base
-gh pr list --head <branch>         # existing PR? (or glab mr list --source-branch)
+git remote get-url origin                    # detect platform (github.com → gh, gitlab.com → glab)
+git rev-parse --abbrev-ref HEAD              # current branch
+git rev-parse --git-dir                      # worktree detection (see above)
+git fetch origin <default-branch> --quiet    # refresh base ref
+git status                                   # working tree state
+git log origin/<default-branch>..HEAD --oneline   # commits ahead of base
+git diff --name-only origin/<default-branch>...HEAD  # files in the PR
+git ls-remote --heads origin <branch>        # branch already on origin?
+gh pr list --head <branch>                   # existing PR? (or glab mr list --source-branch)
 ```
 
-**Stop and ask the user if:**
-- Current branch is `main`/`master` — you cannot PR from main into main. Ask if they want to create a feature branch first. **Never** use `git reset`, `git push --force`, or any history rewriting on main — not even as an "option." If the work is already pushed to main, tell the user the PR opportunity has passed for this change.
-- There are uncommitted changes — ask if they want to commit first or exclude them.
-- There are no commits ahead of base — nothing to PR.
-- There's already an open PR for this branch — switch to **update flow** (see below).
+### Auto-recover: on main/master
+
+If `HEAD` is on `main`/`master`, **do not stop** — recover automatically:
+
+1. Confirm none of the local commits ahead of base have been pushed to `origin/main`. If `git log origin/main..HEAD` is empty AND the working tree is clean, there is nothing to PR — exit. If commits ahead of base have **already been pushed to origin/main**, stop: the PR window has passed (see Hard Stops).
+2. Pick a feature branch name (see [Branch naming](#branch-naming)).
+3. `git checkout -b <name>` — this carries any uncommitted changes onto the new branch and leaves `main` untouched. **Concurrency:** if other runs may share this checkout (autonomous/parallel mode), do NOT `git checkout -b` in place — isolate in a worktree per [Concurrency safety](#concurrency-safety).
+4. If there are uncommitted changes that belong in the PR, `git add` the relevant files (those whose paths overlap the intended PR scope) and commit with a conventional-commit message derived from the diff. Leave unrelated dirty files alone.
+5. Continue with the normal flow.
+
+**Never** run `git reset`, `git push --force`, `git push --force-with-lease`, or any history rewrite on `main`/`master` — not as a recovery, not as an "option," not ever.
+
+### Auto-recover: dirty working tree overlapping the PR diff
+
+If modified/staged files overlap the PR diff, **do not stop** — stage and commit them as part of the PR with a conventional-commit message derived from the changes. Files outside the PR diff stay untouched.
+
+### Hard Stops
+
+These are the *only* cases where you must stop. In autonomous mode, return the question to the parent instead of blocking:
+
+- **Work already pushed to `origin/main`/`origin/master`.** The PR window for those commits has passed. Do not try to "rescue" it with history rewriting. Tell the user.
+- **No commits ahead of base AND no uncommitted changes.** Nothing to PR.
+- **Open PR already exists for this branch.** Switch to **update flow** (see below) — this isn't really a stop, just a branch in logic.
+- **Remote branch diverged.** `git log HEAD..origin/<branch>` is non-empty after fetching. Reconcile by inspecting; never force-push.
+
+**Do NOT block on:** untracked files, or modified files outside the PR diff. These are common in worktrees (test artifacts, scratch logs, leftovers from a parent agent) and are already excluded from `origin/<base>...HEAD`. Note them in your report and proceed.
+
+### Branch naming
+
+Derive deterministically — do not ask:
+
+1. If there are commits ahead of base, parse the latest commit subject. Take its conventional prefix (`feat`, `fix`, `chore`, `docs`, `refactor`) and slugify the rest: `<type>/<kebab-slug>` (max ~50 chars).
+2. Otherwise (only uncommitted changes), pick the prefix from the change shape — `fix:` for bug language in modified code, `docs:` for `.md`-only, `chore:` for config/tooling, else `feat:`. Slug from the most-changed top-level directory or filename stem.
+3. If a branch by that name already exists locally, append `-2`, `-3`, etc.
+4. **Concurrent/autonomous mode:** deterministic slugs collide when parallel runs share a base. Append a unique suffix — the parent-supplied issue/task number if available, else the short HEAD SHA (`git rev-parse --short HEAD`): `<type>/<slug>-<issue-or-sha>`. See [Concurrency safety](#concurrency-safety).
+
+**Branch already on origin, but no open PR** (common after a draft-escalation push from `devpilot-resolve-issues`):
+1. Only enter this branch if `git ls-remote --heads origin <branch>` returned a SHA. If it was empty, skip — `git push -u origin HEAD` will create the remote ref normally.
+2. `git fetch origin <branch>` — see if remote has commits you don't.
+3. If `git log HEAD..origin/<branch>` is non-empty, stop — the remote diverged. Reconcile by inspecting; never force-push.
+4. Otherwise `git push -u origin HEAD` will fast-forward (or be a no-op). Then create the PR normally.
 
 ## Read the Diff (Required)
 
 ```bash
-git diff <base>...HEAD             # full diff
-git log <base>..HEAD --oneline     # commit history
+git diff origin/<default-branch>...HEAD   # full diff (always vs origin, not local)
+git log origin/<default-branch>..HEAD --oneline
 ```
 
 **You MUST read and understand the actual changes before writing the description.**
@@ -103,7 +178,7 @@ If no project template exists, **read ONE template** based on what you found in 
 - Leave "For Reviewers (human)" items unchecked — those are for humans
 - For bug fixes, describe the bug, root cause, and fix separately
 
-**Show the draft to the user before creating or updating.** Let them edit the title and body.
+**Do not stop to "show the draft for approval."** Write the strongest title and body you can from the diff and create the PR directly. Report the URL afterward; the user can `gh pr edit <number>` if they want to tweak. In autonomous mode (invoked by a parent skill), include the final body inline in your response so the parent can surface it.
 
 ## Create and Report
 
@@ -132,7 +207,7 @@ glab mr view <number>              # GitLab
 - **Reviewer feedback** on description → address the specific feedback while keeping the rest.
 - **Draft → Ready** → update description if it was a WIP placeholder, then mark ready.
 
-**4. Show draft to user, then execute:**
+**4. Execute the update directly** (do not stop for approval):
 ```bash
 gh pr edit <number> --title "..." --body "..."    # GitHub
 glab mr update <number> --title "..." --description "..."  # GitLab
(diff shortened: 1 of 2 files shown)


skills/devpilot-pr-creator/SKILL.md
```diff
@@ -0,0 +1,257 @@
...
+## Concurrency safety
+
+This skill can run **several times at once** — a parent skill (`devpilot-resolve-issues`, `devpilot-auto-feature`) may fan out parallel agents that each open a PR. Concurrent runs corrupt each other only when they **mutate shared repo state** in the **same working directory**: `git checkout -b`, staging, and committing all move the shared `HEAD`/index, so two runs racing in one checkout clobber each other's branch and staged files. Reads (`git fetch`, `git diff`, `git log`, `gh pr list`) are safe to overlap.
+
+Follow these rules whenever concurrency is possible (autonomous mode, or the user says runs overlap):
+
+- **Never mutate a working tree you do not own.** If you would need `git checkout`/`git checkout -b`/`git switch` to get onto a branch — i.e. you are on `main`/`master` in a shared checkout — do **not** switch in place. Create an isolated worktree instead and do all mutation there:
+  ```bash
+  git fetch origin <default-branch> --quiet
+  git worktree add -b <branch> ../pr-<branch> origin/<default-branch>   # isolated HEAD + index
```
SiyuQian: ### [Blocking] Worktree based on `origin/<default-branch>` drops the work being PR'd

**Behavior today on this branch:** This recipe is reached from step 3 of "Auto-recover: on main/master" — i.e. exactly when the work exists as local commits ahead of `origin/main` and/or uncommitted changes in the shared checkout. Creating the worktree at `origin/<default-branch>` gives a clean tree at the remote tip: the local commits are not there (they live on local `main`), and git cannot carry a dirty index into a new worktree, so the modified files stay behind. The next line says "then stage/commit/push here", where there is nothing to stage. The run then reads `git log origin/main..HEAD` as empty and a clean `git status`, and falls into the Hard Stop "No commits ahead of base AND no uncommitted changes. Nothing to PR."

**Why that's a problem:** In autonomous mode the agent silently produces no PR (or an empty one) and abandons the user's work in the shared checkout, where a second concurrent run can still clobber it — the opposite of this section's goal. The instruction it replaces is explicitly documented as carrying uncommitted changes; the substitute silently does not. `devpilot-auto-feature` is the concrete victim: it commits each task on the current branch and never creates a branch or worktree, so its PR step always lands on this path.

**Suggested change:** Base the worktree on the current commit — `git worktree add -b <branch> <path> HEAD` — so local commits come along. Uncommitted changes still cannot ride along, so handle that case explicitly: either `git stash push` the scoped paths in the shared checkout and `git stash pop` inside the new worktree (the stash is shared across worktrees), or declare a dirty shared checkout a Hard Stop that returns to the parent. Then reword step 3 of "Auto-recover: on main/master" so the "carries any uncommitted changes" claim attaches only to the `git checkout -b` branch of the fork.

**Confidence:** high

skills/devpilot-pr-creator/SKILL.md
```diff
@@ -0,0 +1,257 @@
...
+  cd ../pr-<branch>                                                     # then stage/commit/push here
+  ```
+  If you were **already handed a dedicated branch/worktree** (the common autonomous case — you are on a feature branch in a linked worktree), you already own it; stage and commit in place, no new worktree needed.
+- **Make the branch name collision-proof.** Deterministic naming (below) makes two parallel runs on similar diffs pick the *same* name and collide on push. In concurrent/autonomous mode, append a short unique suffix to the derived slug — the issue/task number when the parent provides one, otherwise the short HEAD SHA: `<type>/<slug>-<issue-or-sha>`. Never reuse `Date.now()`-style timestamps.
+- **Let push, not a pre-check, arbitrate the race.** Two runs may both pass `git ls-remote` and then both push the same ref. Rely on the non-fast-forward push *rejection* as the signal — if `git push -u origin HEAD` is rejected, re-fetch and inspect per [Hard Stops](#hard-stops); never force-push to resolve it.
+- **Clean up an isolated worktree** you created once the PR is open: `git worktree remove ../pr-<branch>` (skip if it still holds uncommitted work).
```
SiyuQian: ### [Should-fix] `../pr-<branch>` breaks on slashed branch names and contradicts the repo's own worktree scheme

**Behavior today on this branch:** "Branch naming" always yields `<type>/<slug>`, so this expands to `../pr-fix/some-slug`. Reproduced on git 2.50.1: `git worktree add` silently `mkdir -p`s a `pr-fix/` directory beside the repo root, and then this cleanup command, run from inside the worktree the recipe told the agent to `cd` into, fails with `fatal: '../pr-fix/some-slug' is not a working tree` because `..` now resolves to `pr-fix/`. With a flat branch name the removal succeeds but leaves cwd inside a deleted directory, so every later command in the run errors. The repo already settled a different scheme in `devpilot-resolve-issues/references/worktree-management.md`: `<repo-root>.worktrees/<name>`, chosen so that "`rm -rf <repo>.worktrees` cleans every dangling worktree at once" and "the dir name embeds the repo name" — neither property holds for a bare `../` sibling, and `.gitignore` covers only `.worktrees/`.

**Why that's a problem:** Cleanup never fires, so every isolated run leaks an orphan worktree pinning a branch plus a stray directory outside the repo; the next run then fails on "branch already exists". Worse for a concurrency fix: `../pr-<slug>` created from two different repos under the same parent is the same directory, so this introduces a fresh cross-repo collision.

**Suggested change:** Reuse the established scheme rather than restating a second layout — `WORKTREE="$(git rev-parse --show-toplevel).worktrees/pr-$(echo <branch> | tr / -)"`, the same expression used at `worktree-management.md:68`. Capture `MAIN="$(git rev-parse --show-toplevel)"` before the `cd`, and `cd "$MAIN"` before removing. Use that absolute variable in both the `git worktree add` on the line above and the removal here, and link to `worktree-management.md` → Cleanup for the `clean -fdX` ladder instead of the bare one-liner.

**Confidence:** high

skills/devpilot-pr-creator/SKILL.md
```diff
@@ -0,0 +1,257 @@
...
+
+- **Never mutate a working tree you do not own.** If you would need `git checkout`/`git checkout -b`/`git switch` to get onto a branch — i.e. you are on `main`/`master` in a shared checkout — do **not** switch in place. Create an isolated worktree instead and do all mutation there:
+  ```bash
+  git fetch origin <default-branch> --quiet
+  git worktree add -b <branch> ../pr-<branch> origin/<default-branch>   # isolated HEAD + index
+  cd ../pr-<branch>                                                     # then stage/commit/push here
+  ```
+  If you were **already handed a dedicated branch/worktree** (the common autonomous case — you are on a feature branch in a linked worktree), you already own it; stage and commit in place, no new worktree needed.
+- **Make the branch name collision-proof.** Deterministic naming (below) makes two parallel runs on similar diffs pick the *same* name and collide on push. In concurrent/autonomous mode, append a short unique suffix to the derived slug — the issue/task number when the parent provides one, otherwise the short HEAD SHA: `<type>/<slug>-<issue-or-sha>`. Never reuse `Date.now()`-style timestamps.
```
SiyuQian: ### [Should-fix] Short HEAD SHA is identical across exactly the parallel runs it must disambiguate

**Behavior today on this branch:** The fallback suffix is "the short HEAD SHA (`git rev-parse --short HEAD`)". But the hazard this section defines is N runs racing "in the **same working directory**" on `main` in "a shared checkout" — every one of them reads the same `HEAD`, so the suffix is byte-identical across all of them. Combined with branch-naming rule 2 (slug derived from the same shared dirty tree), two concurrent runs still derive the same branch name, still target the same worktree path, and the second `git worktree add -b` fails on "branch already exists" before the push race can arbitrate anything. Rule 3's `-2`/`-3` local-existence dedupe is itself racy, and its ordering relative to this rule is unspecified. The issue-number alternative is currently unreachable: the only parent with an issue number, `devpilot-resolve-issues`, hands over an already-created branch and so takes the "you already own it" path without ever deriving a name. The one source that is genuinely unique per run — a timestamp — is banned in the same sentence.

**Why that's a problem:** The second of the two failure modes this PR claims to fix is unfixed in the precise case it targets, now with a false sense of safety.

**Suggested change:** Require a per-run token that differs between concurrent agents rather than one derived from shared state — the agent/task id the parent hands down, or a nonce (`openssl rand -hex 2`, `$$`). Keep the HEAD SHA only as an extra component, never as the sole discriminator. If the timestamp ban exists for retry-idempotency, say so and pair it with "reuse the branch name recorded earlier in this run". Also state whether this rule applies before or after rule 3's `-2`/`-3` dedupe. Same rule restated at line 113 needs the same fix.

**Confidence:** high

skills/devpilot-pr-creator/SKILL.md
```diff
@@ -0,0 +1,257 @@
...
+
+This skill can run **several times at once** — a parent skill (`devpilot-resolve-issues`, `devpilot-auto-feature`) may fan out parallel agents that each open a PR. Concurrent runs corrupt each other only when they **mutate shared repo state** in the **same working directory**: `git checkout -b`, staging, and committing all move the shared `HEAD`/index, so two runs racing in one checkout clobber each other's branch and staged files. Reads (`git fetch`, `git diff`, `git log`, `gh pr list`) are safe to overlap.
+
+Follow these rules whenever concurrency is possible (autonomous mode, or the user says runs overlap):
+
+- **Never mutate a working tree you do not own.** If you would need `git checkout`/`git checkout -b`/`git switch` to get onto a branch — i.e. you are on `main`/`master` in a shared checkout — do **not** switch in place. Create an isolated worktree instead and do all mutation there:
+  ```bash
+  git fetch origin <default-branch> --quiet
+  git worktree add -b <branch> ../pr-<branch> origin/<default-branch>   # isolated HEAD + index
+  cd ../pr-<branch>                                                     # then stage/commit/push here
```
SiyuQian: ### [Should-fix] `cd` here contradicts the file's own "the parent owns cwd" invariant

**Behavior today on this branch:** Line 42, in "Worktree / autonomous invocation", states unconditionally: "Do not `cd` out of the worktree; the parent owns cwd." This section applies "whenever concurrency is possible (autonomous mode, or the user says runs overlap)" and issues an unconditional `cd` with no instruction to restore cwd before returning control. The trailing paragraph carves out only the "already handed a dedicated branch/worktree" case; the on-main-in-a-shared-checkout case still tells an autonomous run to change the cwd the parent owns.

**Why that's a problem:** Two directly opposed orders in one document, and an agent will pick either. The contract is enforced elsewhere too — `devpilot-resolve-issues/references/subagent-spec.md:16` tells child agents "Do not `cd` outside the worktree, do not run `git worktree add`/`remove`/`prune` … The controller owns worktree lifecycle" — so an agent obeying this line inside a resolve-issues run can remove a worktree the controller still tracks. `devpilot-auto-feature` does no cwd management after its PR step, so a run ending here (or in a directory it just deleted) leaves the parent operating from the wrong tree.

**Suggested change:** Drop the `cd` entirely and use `git -C "$WORKTREE" ...` for the stage/commit/push steps, which needs no cwd change at all. If you keep it, scope line 42 explicitly ("do not `cd` out of a worktree the parent handed you") and require capturing `MAIN="$(git rev-parse --show-toplevel)"` before the `cd` plus `cd "$MAIN"` before returning. Cross-reference `subagent-spec.md:16` so the parent/child contracts agree.

**Confidence:** high

.claude/skills/devpilot-pr-creator/SKILL.md
```diff
@@ -0,0 +1,257 @@
...
+2. Pick a feature branch name (see [Branch naming](#branch-naming)).
+3. `git checkout -b <name>` — this carries any uncommitted changes onto the new branch and leaves `main` untouched. **Concurrency:** if other runs may share this checkout (autonomous/parallel mode), do NOT `git checkout -b` in place — isolate in a worktree per [Concurrency safety](#concurrency-safety).
+4. If there are uncommitted changes that belong in the PR, `git add` the relevant files (those whose paths overlap the intended PR scope) and commit with a conventional-commit message derived from the diff. Leave unrelated dirty files alone.
+5. Continue with the normal flow.
+
+**Never** run `git reset`, `git push --force`, `git push --force-with-lease`, or any history rewrite on `main`/`master` — not as a recovery, not as an "option," not ever.
+
+### Auto-recover: dirty working tree overlapping the PR diff
+
+If modified/staged files overlap the PR diff, **do not stop** — stage and commit them as part of the PR with a conventional-commit message derived from the changes. Files outside the PR diff stay untouched.
```
SiyuQian: ### [Should-fix] Syncing the mirror activates auto-commit, which the repo's `CLAUDE.md` forbids

**Behavior today on this branch:** This rule ("do not stop — stage and commit them as part of the PR") and the "Operating mode: automatic by default" paragraph above it already existed in `skills/`, but the installed mirror was stale since PR #92/#98 and did not carry them. This PR's sync brings them live in `.claude/skills/`, which is the tree Claude Code actually loads for this repo. Root `CLAUDE.md` (and the identical `AGENTS.md`), under "Safety rules the harness can't enforce", states: "Never commit without an explicit user ask." Nothing in the diff carves out an exception. Flagging it here rather than as a pre-existing issue because this PR is what makes the instruction reachable for the first time.

**Why that's a problem:** As installed, any `/pr` run in this repo may now commit dirty files the user never asked to commit, which is the one thing the repo-wide convention file says never to do without an ask. `CLAUDE.md` is repo-wide and higher-priority, so an agent gets contradictory instructions with real write consequences.

**Suggested change:** Resolve the conflict in one direction explicitly. Either add the carve-out to `CLAUDE.md` ("except when `devpilot-pr-creator` runs in autonomous mode"), or scope the auto-commit in the source `skills/devpilot-pr-creator/SKILL.md` to autonomous/parent-invoked runs only, leaving human-invoked runs to ask. Edit `skills/` and re-run `make sync-skills` — do not hand-patch this mirror.

**Confidence:** medium — the conflict text is literal in both files, but whether `CLAUDE.md`'s rule is intended to bind skill-internal autonomous flows is an author call.

SiyuQian (commented):
@SiyuQian — 1 blocking, 4 should-fix · reviewed at fc71189

## PR Review: fix(pr-creator): make skill safe to run concurrently — #157

### Verdict
**Ready to merge:** No
The isolation recipe that is the centrepiece of this change creates the worktree from `origin/<default-branch>`, so on the one path it is written for (auto-recover on `main`) it silently leaves behind the very commits and dirty files the run exists to ship.

### TL;DR
Fix the worktree base first: `git worktree add -b <branch> <path> origin/<default-branch>` starts from the remote tip, so local commits ahead of base and uncommitted changes never make it into the isolated tree, and the run falls through to "Nothing to PR". Three supporting rules (path scheme, SHA suffix, `cd`) also do not hold up in the exact concurrent scenario the section defines.

### Strengths
- The diagnosis at [line 48](https://github.com/SiyuQian/devpilot/blob/fc71189d478800d433ddf9802db04d9815c46db7/skills/devpilot-pr-creator/SKILL.md#L47-L49) is precise about the real mechanism — shared `HEAD`/index in one working directory — and explicitly marks reads (`git fetch`, `git diff`, `gh pr list`) as safe to overlap, which stops the rule from over-applying.
- ["Let push, not a pre-check, arbitrate the race"](https://github.com/SiyuQian/devpilot/blob/fc71189d478800d433ddf9802db04d9815c46db7/skills/devpilot-pr-creator/SKILL.md#L59-L61) is the correct call: it rejects a TOCTOU `git ls-remote` pre-check and reinforces the never-force-push prohibition instead of carving an exception into it.
- The `.claude/skills/` mirror is byte-identical to `skills/` at head (verified by `diff`), matching the `make sync-skills` direction documented in `ARCHITECTURE.md`, so the larger mirror hunk is a genuine backfill rather than divergent policy.

### Unknown-Unknowns Sweep
- Local pattern fit: the new section re-defines autonomous mode (already established at line 39) and introduces three spellings of the trigger condition ("concurrency is possible", "concurrent/autonomous mode", "autonomous/parallel mode"); it also adds no row to the file's "Common Mistakes" table, which every prior behavioural rule in this file has.
- Blast radius: grep-only fallback (code graph skipped — docs-only Markdown diff, no code symbols). Two parent skills consume this contract: `devpilot-resolve-issues` hands over a dedicated branch in a linked worktree and takes the safe "you already own it" path; `devpilot-auto-feature` commits on the current branch and hands over nothing, so it lands squarely on the broken worktree recipe.
- Known pitfalls (concurrency + reversibility): both classic failure shapes are present — an isolation primitive that silently changes the data it isolates, and a "unique" suffix derived from state the racing parties share.
- Stale-training check: git behaviour was verified empirically on git 2.50.1 rather than asserted — `worktree add` does `mkdir -p` intermediate dirs for a slashed path, and `worktree remove` of the worktree you are standing in succeeds while leaving cwd deleted.
- Hand-rolled vs. off-the-shelf: the repo already owns a worktree lifecycle spec at `skills/devpilot-resolve-issues/references/worktree-management.md` (path scheme, `git clean -fdX` before removal, `--force` reserved as last resort). The new section reinvents a weaker version instead of linking to it; reusing it resolves the path and cleanup findings for free.

Plus 6 lower-priority design notes that fell below the reporting threshold: the race loser dead-ends into Hard Stops with no rename-and-retry path; the cleanup one-liner skips the established `clean -fdX` ladder; the 17-line section arguably belongs in `references/` per `GOLDEN_PRINCIPLES.md` #14; no "Common Mistakes" row was added; three competing names for one mode with "shared checkout" left undefined; and the section restates the autonomous-mode definition without adding a preflight ownership check.

### Security / Performance coverage
8/8 security required-checks completed · 7/7 performance required-checks completed
- `authn_vs_authz`, `error_response_leakage`, `cross_host_credential_propagation`, `tls_transport`: not applicable — no endpoint, handler, RPC, or outbound call in the diff; `git`/`gh` use existing credential helpers.
- All 7 performance items except `serial_independent_awaits`: not applicable — Markdown instruction file with no request path, DB schema, or hot loop.
- Assumptions recorded: 2 — `<branch>` is derived from the local commit subject by the agent itself, so the path-shaped interpolation at line 55 is a correctness bug rather than an injection vector; and `git worktree add` refuses a non-empty target, so a pre-existing sibling directory aborts the run rather than being clobbered.
- Dependency reality: agent_f skipped (no dependency manifest touched).

### Inline findings (count by severity)
- Blocking: 1
- Should-fix: 4

### Open Questions
- The GitHub diff shows **515 files / +73,108** because the branch was cut from merge-base `2ce76b0` and replayed 241 commits of `main`'s history; the real delta against `origin/main` is 2 files / +125/−15. Repo settings are squash-only (`allow_merge_commit:false`, `allow_rebase_merge:false`), so `main` will not inherit duplicate history — but the diff is unreviewable as presented. Rebase onto `origin/main` before marking ready?
- The premise names `devpilot-resolve-issues` and `devpilot-auto-feature` as parents that "fan out parallel agents that each open a PR". Neither does today — `devpilot-resolve-issues` explicitly forbids parallel implementers on the same branch, and `devpilot-auto-feature` is sequential with a single PR. Is the parent-side change landing separately, or should the premise be softened to cover two human sessions sharing one checkout?
- No parent currently passes an issue/task number down to this skill, so the "issue/task number when the parent provides one" branch of the suffix rule is unreachable in practice. Intended for a future parent change?
- This PR is still marked draft — reviewed on request.

---

> _Automated review. Code graph / AST facts: not available (docs-only Markdown diff; no code symbols to index). Verify findings before acting; not a substitute for human review._

skills/devpilot-pr-creator/SKILL.md
SiyuQian: Fixed in 732ba67. The worktree is now created off `HEAD` (`git worktree add -b "$BRANCH" "$WORKTREE" HEAD`) with an inline comment saying why, so local commits ahead of base come along. A dirty shared checkout is now an explicit Hard Stop that returns to the parent rather than creating the worktree and leaving the work behind, and the text states that only `git checkout -b` in a checkout you own carries uncommitted changes.

skills/devpilot-pr-creator/SKILL.md
SiyuQian: Fixed in 732ba67. Switched to the established scheme: `WORKTREE="$(git rev-parse --show-toplevel).worktrees/pr-$(printf %s "$BRANCH" | tr / -)"`, with the slashes flattened so `<type>/<slug>` names work. Both the `git worktree add` and the `git worktree remove` now use that absolute `$WORKTREE`, and the text links to `worktree-management.md` for the cleanup ladder and calls out why a `../` sibling is wrong (cross-repo collision, escapes `.gitignore`).

skills/devpilot-pr-creator/SKILL.md
SiyuQian: Fixed in 732ba67. The fallback is now a per-run token — parent-supplied issue/task number, else parent-supplied agent/task id, else a nonce (`openssl rand -hex 2`) — and the text explicitly rejects the short HEAD SHA as a sole discriminator for the shared-checkout case. Ordering against rule 3 is now stated (this suffix replaces the racy `-2`/`-3` dedupe), and the timestamp ban is justified by retry-idempotency with "record the suffix once and reuse it on retry". Rule 4 under Branch naming carries the same wording.

skills/devpilot-pr-creator/SKILL.md
SiyuQian: Fixed in 732ba67. The `cd` is gone; the recipe drives the worktree with `git -C "$WORKTREE"` and says "never cd", cross-referencing both line 42 ("the parent owns cwd") and `subagent-spec.md`. Cleanup notes that cwd is still valid precisely because nothing ever cd`d.

.claude/skills/devpilot-pr-creator/SKILL.md
SiyuQian: Not fixed — leaving this open deliberately. Both suggested directions (add a carve-out to `CLAUDE.md`/`AGENTS.md`, or scope the auto-commit in `skills/` to autonomous runs only) change repo-wide policy or skill behavior, which is an author decision this unattended pass will not guess at. Flagging it as the remaining blocker for this PR. Note also that `make sync-skills` cannot be used here: main`s mirror is stale for many other skills, so running it pulls ~20 unrelated files into this PR and deletes the `devpilot-learn`/`devpilot-resolve-issues` symlinks. The pr-creator mirror was synced by hand to stay in scope.

SiyuQian (commented):
## Review — `732ba67` (devpilot:babysit-prs unattended pass)

**Verdict: 1 blocking issue remaining (author decision required).**

### What this pass changed

The PR was `CONFLICTING` at discovery with an effective diff of **515 files / +73k lines** — `main`'s history had been rewritten after this branch was cut, so the merge-base was an ancient commit and the PR was replaying 241 duplicate commits. Merging `origin/main` into the branch (no rebase, no force-push) collapsed the PR to its actual intent:

```
.claude/skills/devpilot-pr-creator/SKILL.md | 122 ++++++++++++++++++----
skills/devpilot-pr-creator/SKILL.md         |  22 ++++-
```

The add/add conflict on both copies was resolved by keeping `main`'s canonical skill text plus this PR's `Concurrency safety` section, then mirroring it into the installed copy — which is the PR's stated intent and matches the repo's mirror convention.

Then `732ba67` addressed four of the five open review threads:

| Thread | Fix |
|---|---|
| [Blocking] worktree drops the work being PR'd | worktree now created off `HEAD`, not `origin/<default>`; dirty shared checkout is an explicit Hard Stop |
| `../pr-<branch>` layout | now `<repo-root>.worktrees/pr-<flattened-branch>`, absolute, used by both add and remove |
| short HEAD SHA as suffix | replaced with a per-run token (issue/task number → agent id → nonce); ordering vs rule 3 stated |
| `cd` contradicts parent-owns-cwd | `cd` removed entirely; the recipe uses `git -C "$WORKTREE"` |

### Remaining blocker

**`.claude` mirror sync activates auto-commit, which `CLAUDE.md` forbids** (thread left open). Both proposed resolutions — carving an exception into `CLAUDE.md`/`AGENTS.md`, or scoping the auto-commit in `skills/` to autonomous runs only — are repo-policy or behavior decisions. An unattended pass will not guess at them, so this PR stays a draft and is **not** labeled `ready to merge`.

### Note on `make sync-skills`

Do not use it to sync this mirror. `main`'s `.claude/skills/` is stale for ~20 other files, and `rsync -a --delete` also removes the `devpilot-learn` and `devpilot-resolve-issues` symlinks. The pr-creator mirror was synced by hand to keep the diff in scope; the broader drift deserves its own PR.

### Gates on `732ba67`

- `go build ./...` — pass
- `go test ./...` — pass
- `gofmt -l .` — clean
- Mirror is byte-identical to `skills/` — verified
- CI (`lint`, `test`) — passed on `8d4e880`; re-running on `732ba67`

*Scope classification: Documentation-only (two Markdown skill-instruction files). Merge withheld on the open thread, not on scope.*

